PR #134, opened 2015-03-01
Notification Model String Hash

yoyos:
I added a compatibility for ActiveAdmin which works with JSON as strings

Problem was I had always the Must be a Hash message because I was sending a string from a textfield.

### DIFF
--- a/lib/rpush/client/active_record/notification.rb
+++ b/lib/rpush/client/active_record/notification.rb
@@ -21,7 +21,7 @@ module Rpush
 
         def data=(attrs)
           return unless attrs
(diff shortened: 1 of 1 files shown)


ileitch:
This sounds like an issue with ActiveAdmin rather than Rpush. Why is it trying to assign a string to a JSON attribute?

yoyos:
Didn't resolve the problem. the issue was something else. Detailed here: https://github.com/rpush/rpush/issues/141

yoyos:
I still need this change.
The :data field is saved as a string in db so it is abnormal we cannot input a string.

PR #137, opened 2015-03-11
Add WNS support (Windows RT)

voidless:
There are no spec tests, but it works with Asus tablet
Only toast type is supported, but other types are easy to add


ileitch:
This really needs some test coverage before I can merge. Should be pretty easy to add given the similarity with WPNS.

matiaslina:
I will work on the test coverage for this pr :)

matiaslina:
Here's the [patch](https://gist.github.com/matiaslina/5b3d90381f4017d7faec) based on the master branch of @voidless. 

You can apply the patch after the merge of this pr, or voidless can merge it in his master branch.

Issue #141, opened 2015-03-31
bad behavior with json notification :data field and ActiveAdmin


yoyos:
Hello

I posted a bug report https://github.com/activeadmin/activeadmin/issues/3871 for a bug I thought was coming from ActiveAdmin but it seems rpush changed a default behavior on the :data field.

For example, If I totally comment out these lines in file `lib/rpush/client/active_record/notification.rb`, ActiveAdmin can display json strings correctly in its edit form like `{"foo":"bar"}` and not `{"foo"=>"bar"}` anymore : 

```
def data=(attrs)
   return unless attrs
   attrs = JSON.parse(attrs) if attrs.is_a?(String)
   fail ArgumentError, "must be a Hash" unless attrs.is_a?(Hash)
   write_attribute(:data, multi_json_dump(attrs.merge(data || {})))
end
```

Are these lines really useful ? Could you make it work with activeadmin please ? If possible, is there a way to do it without changing default behaviors if not necessary ?

Thanks !

aried3r:
Hey! Thanks for reporting the issue! Since there haven't been any new feedback for some time I'm closing this issue. Please feel free to open a new issue if you're still experiencing this issue with the latest Rpush and ActiveAdmin release.

PR #143, opened 2015-04-04
ActiveAdmin support

yoyos:
I'm trying to solve this issue: https://github.com/rpush/rpush/issues/141
I think rpush should accept and parse strings for data and registration_ids fields because they are saved as string in db. We just have to respect the final format.

This lets ActiveAdmin works correctly as they can treat these fields like simple strings.

I'm new to code sharing on github with pull requests, unit testing etc.
Please don't hit me and guide me ^^

### DIFF
--- a/lib/rpush/client/active_record/notification.rb
+++ b/lib/rpush/client/active_record/notification.rb
@@ -21,17 +21,26 @@ module Rpush
 
         def data=(attrs)
           return unless attrs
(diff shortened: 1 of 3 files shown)


yoyos:
Here is what I did in a few minutes with these changes

capture d ecran 2015-04-05 01 59 00

marcolong:
Any news?

PR #163, opened 2015-06-03
Add alternate ActiveRecord Daemon store for use with Oracle

latortuga:
Oracle does not allow an ORDER clause when using FOR UPDATE so this store achieves the same locking without ordering in SQL.

For lack of a config variable, this custom store can be activated by placing the following at the end of the rpush initializer provided you are already using ActiveRecord client.

```
require 'rpush/daemon/store/oracle'
Rpush::Daemon.store = Rpush::Daemon::Store::Oracle.new
```

### DIFF
--- a/lib/rpush/daemon/store/oracle.rb
+++ b/lib/rpush/daemon/store/oracle.rb
@@ -1,0 +1,29 @@
+require 'rpush/daemon/store/active_record'
+
+# The default AR implementation throws an error when trying to lock resources
+# when using an ORDER clause with FOR UPDATE. Instead of ORDER in SQL, we
+# simply sort the result set in memory.
+module Rpush
+  module Daemon
+    module Store
+      class Oracle < Rpush::Daemon::Store::ActiveRecord
+        def deliverable_notifications(limit)
+          with_database_reconnect_and_retry do
+            Rpush::Client::ActiveRecord::Notification.transaction do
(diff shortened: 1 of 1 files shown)


ileitch:
The order is used only to ensure that the oldest notifications are delivered first. Sorting after they've been loaded probably doesn't achieve anything.

latortuga:
You're absolutely right to the extent that the oldest ones won't be
selected anymore. However, it will still ensure that the oldest ones in the
result set are delivered first. Happy to fix if you'd rather do without it.

We are running this patch with the latest rpush in staging at the moment
and it is working great.

ileitch:
It's probably not needed, but it's safer to keep it this way as something somewhere may assume the notifications are in order.

Thanks!

ileitch:
Actually, in hindsight I think a better approach is to have the ActiveRecord store transparently support Oracle without the need for a separate store....

ileitch:
Can you give me the output from `Hash[ActiveRecord::Base.configurations[Rails.env].map { |k,v| [k.to_sym,v] }][:adapter]` for Oracle?

latortuga:
```ruby
2.1.5 :002 >  Hash[ActiveRecord::Base.configurations[Rails.env].map { |k,v| [k.to_sym,v] }][:adapter]
 => "oracle_enhanced"
```

PR #166, opened 2015-06-09
Replace find_by_name with find_by in README

dzjuck:
`find_by_name` works with ActiveRecord backend and does not with Mongoid and Redis backends.
`find_by` works with ActiveRecord and Mongoid backends, and does not with Redis (looks like it's not difficult to implement, will try to do it). 

Hope it will improve readme for a little. Thank you.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ app.save!
 
 ```ruby
 n = Rpush::Apns::Notification.new
-n.app = Rpush::Apns::App.find_by_name("ios_app")
+n.app = Rpush::Apns::App.find_by(name: "ios_app")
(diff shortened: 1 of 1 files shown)


ileitch:
Find on Redis can only done if that field has an index, which name does. However, Modis might not implement find_by... I don't remember :)

dzjuck:
I'd tried, it does not :) If i understand right, rpush create index for name field and Modis can call 
```ruby
where(name: 'test')
``` 
So it's possible to implement find_by in Modis as alias for
```ruby
where(...).first
```
I'm not familiar with Redis operations. Does it have better when smembers operator for finding first match?

PR #185, opened 2015-07-28
Fix typo in Oracle support

latortuga:


latortuga:
I'm confident in this change because I put the following patch at the end of my Rpush initializer and everything is working with Oracle:

```ruby
if Rpush::VERSION == '2.5.0'
  require 'rpush'
  require 'rpush/daemon'
  require 'rpush/daemon/store/active_record'

  module Rpush
    module Daemon
      module Store
        class ActiveRecord
          private

          def claim(relation)
            notifications = relation.lock(true).to_a
            @using_oracle ? notifications.sort_by(&:created_at) : notifications
          end
        end
      end
    end
  end
end
```

ileitch:
:scream: oops!

PR #187, opened 2015-08-07
not able to stop rpush sometimes

zhengxu001:
>>SIGTERM - This signal requests a process to stop running. This signal can be ignored. The process is given time to gracefully shutdown. When a program gracefully shuts down, that means it is given time to save its progress and release resources. In other words, it is not forced to stop. SIGINT is very similar to SIGTERM.

So, "Process.kill('TERM', pid)" sometimes failed and then code will do the loop all the time.

We met this problem for several times during deployment.

### DIFF
--- a/lib/rpush/cli.rb
+++ b/lib/rpush/cli.rb
@@ -35,15 +35,10 @@ module Rpush
 
       STDOUT.write "* Stopping Rpush (pid #{pid})... "
       STDOUT.flush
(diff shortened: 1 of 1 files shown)


ObjColumnist:
I have also seen this on Heroku when redeploying an app

aried3r:
@zhengxu001 Did 2.6.0 solve this for you?

PR #190, opened 2015-08-18
Remove `param` tag from WNS message.

andrey-skat:
Toast does not show if `param` tag passed. Also it does not exist in documentation https://msdn.microsoft.com/en-us/library/windows/apps/br230849.aspx


andrey-skat:
Someone broke tests on jruby 3 months ago...

ileitch:
Thanks!

jkasten2:
There is a launch attribute string on the toast tag that parameters will work for.

https://msdn.microsoft.com/en-us/library/windows/apps/br230846.aspx

```ruby
"<toast launch='#{param}'>
```

PR #192, opened 2015-08-25
Fixed Wns response headers parser

krakatoa:
This solves an issue with the Wns response parsing: the keys in the header response are in downcases.
thanks


ileitch:
This is still fragile, all keys should be explicitly down-cased before use.

krakatoa:
You're totally right. Sent a new commit to be sure it won't fail with future case changes

PR #201, opened 2015-09-24
adding content_available to gcm

victorlcampos:
Hi,
GCM has a especification to use it with IOS that gem is not following.

https://developers.google.com/cloud-messaging/ios/client

Could I implement this?
Here my suggestion, how could i improve this?


aried3r:
Why close this?

PR #204, opened 2015-10-05
Adding content_available to gcm

victorlcampos:
Hi,
to use GCM for IOS it's necessary set content_available in body message. I make this changes, but I don't know how to make the tests pass (how to set up the relational databases and mongo).

Source: https://developers.google.com/cloud-messaging/ios/client

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -244,3 +244,6 @@ DEPENDENCIES
   sqlite3
   stackprof
   timecop
+
+BUNDLED WITH
+   1.10.3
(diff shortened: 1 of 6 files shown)


aried3r:
Thanks for doing this, just stumbled upon this since I wanted to set the same flag myself, but couldn't.

I think you might have to add the migration here: https://github.com/rpush/rpush/blob/357ef6e5329b53229a5c3a53c6ce1ac6f84164de/lib/generators/rpush_migration_generator.rb#L41-L42

And call it `2_5_1` maybe, or w/e version is the next release. But `2_1_1` is misleading.

PR #210, opened 2015-10-21
Update WNS instructions with missing params

pke:
Set the `client_id` and `client_secret` where not set. Without them the sample could would not have worked.


pke:
@ileitch Wow that was quick :)

Issue #224, opened 2015-12-05
`rpush status` doesn't get PID file from Rpush.config


nahaylo:
I changed default path to PID file, e.g: config.pid_file = 'tmp/pids/rpush.pid'

It allows to start 
```
rpush start
* Booting Rails 'development' environment... ✔
```

and stop
```
rpush stop
* Booting Rails 'development' environment... ✔
* Stopping Rpush (pid 27478)... ✔
```

But getting status of process fails
```
rpush status
* Rpush isn't running? /tmp/rpush.pid does not exist.
```

nahaylo:
After some investigation I've found the reason.
It causes because `start` and `stop` load the rails environment, but status isn't.
Fix should be simple, I'm going to create a pull-request.

huteri:
Are you sure you got it right? because I got this error 
```
/Users/huteri/.rvm/gems/ruby-2.1.2/bundler/gems/rpush-aee00e0e2a82/lib/rpush/daemon/rpc/client.rb:6:in `initialize': No such file or directory - connect(2) for "/tmp/rpush.31718.sock" (Errno::ENOENT)
```
when trying to rpush status -e production, and with rpush start -e production

nahaylo:
@huteri 
just checked it works for me

```
push status -e production
* Booting Rails 'production' environment... ✔
* Rpush isn't running? /tmp/pids/rpush.pid does not exist.
```

BTW: It seems that my pull-request is not green, need to fix test :)

huteri:
@nahaylo try to start the rpush first in production then execute rpush status -e production again

nahaylo:
@huteri ok, I've got what you mean.
The problem is that 
```
rpush start -e production
* Booting Rails 'production' environment... ✔
```
didn't return correct status code. For some reason (in my case I simply don't have properly configured production environment, and I think the same for in your case), it didn't started the daemon and didn't reported that.
You can check this, simple start and than try to stop, you will receive the same error. 
Or, run this

```
rpush start
* Booting Rails 'development' environment... ✔
# and then
ps auwx | grep rpush
vitaliy         44039   0,0  0,0  2423372    228 s002  R+    2:51     0:00.00 grep rpush
vitaliy         43977   0,0  0,7  2595356  59532   ??  S     2:51     0:00.18 rpush | 0 queued | 4 dispatchers 
```

and
```
rpush start -e production
* Booting Rails 'production' environment... ✔
# and then
ps auwx | grep rpush
[...]

PR #225, opened 2015-12-05
Fix issue with custom PID file set in Rpush.config

nahaylo:
More details here https://github.com/rpush/rpush/issues/224


latortuga:
LGTM!

PR #226, opened 2015-12-14
Update CI to build on latest JRuby

latortuga:


aried3r:
Not that these are running, but good to see everything JRuby fails ;p

PR #221, opened 2015-11-27
Add content_available to GCM notifications

aried3r:
Sorry for the two pull requests, now from the correct branch.

Adds `content_available` to GCM notifications. Read more about it [here](https://developers.google.com/cloud-messaging/ios/client#receive_messages_through_the_gcm_apns_interface)

PR #227, opened 2015-12-16
Change GCM server URL

aried3r:
Taken from https://developers.google.com/cloud-messaging/http


aried3r:
@victorlcampos Since you seem to be using GCM as well, could you take a look at this one and maybe #221?

victorlcampos:
@aried3r, could it wait for weekend?

aried3r:
Sure

Issue #217, opened 2015-11-07
Rpush Priority High On GCM


huteri:
Hi, I've set the rpush.priority = "high" in gcm, but it seems it does not work, I don't see the parameter in the response result. How do you set this?

schmidp:
in the current master branch, priority on GCM is not supported.
I just added in my branch: https://github.com/openresearch/rpush/commit/ff0ce6df8c2e27981ad01c85fdf90e4ce8ae31a6

huteri:
Hi @schmidp, thank you, this library needs people like you..

PR #228, opened 2015-12-16
Support for High Priority in GCM

huteri:
Hi, currently rpush does not support priority in GCM, and this pull request will make it possible. Referring to #217

### DIFF
--- a/lib/rpush/client/active_model/gcm/notification.rb
+++ b/lib/rpush/client/active_model/gcm/notification.rb
@@ -6,6 +6,7 @@ module Rpush
           def self.included(base)
             base.instance_eval do
               validates :registration_ids, presence: true
+              validates :priority, inclusion: { in: Rpush::Client::ActiveModel::Apns::Notification::APNS_PRIORITIES }, allow_nil: true
 
               validates_with Rpush::Client::ActiveModel::PayloadDataSizeValidator, limit: 4096
               validates_with Rpush::Client::ActiveModel::RegistrationIdsCountValidator, limit: 1000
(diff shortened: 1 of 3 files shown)


lib/rpush/daemon/apns/feedback_receiver.rb
```diff
@@ -45,7 +45,7 @@ def stop
           @thread.join if @thread
         rescue StandardError => e
           log_error(e)
-          reflect(:error, e)
+          reflect(:error, e, @app)
```
aried3r: Why was this change necessary?

lib/rpush/client/active_model/gcm/notification.rb
```diff
@@ -22,6 +23,14 @@ def as_json
             }
             json['collapse_key'] = collapse_key if collapse_key
             json['time_to_live'] = expiry if expiry
+
+            # see https://developers.google.com/cloud-messaging/http-server-ref
+            # GCM also supports APNS and says normal == 5 and high == 10
+            # so we reuse the APNS priority here
+            if priority and priority == Rpush::Client::ActiveModel::Apns::Notification::APNS_PRIORITY_IMMEDIATE
```
aried3r: I think the rest of the app uses `&&` and the `and` and `or` keywords are discouraged by the ruby style guide. https://github.com/bbatsov/ruby-style-guide#no-and-or-or

lib/rpush/client/active_model/gcm/notification.rb
```diff
@@ -6,6 +6,7 @@ module Notification
           def self.included(base)
             base.instance_eval do
               validates :registration_ids, presence: true
+              validates :priority, inclusion: { in: Rpush::Client::ActiveModel::Apns::Notification::APNS_PRIORITIES }, allow_nil: true
```
aried3r: Personally, I'd not reuse the APNS priority, because while `normal` and `high` correspond to `5` and `10` in iOS, this might change. I'd rather have `GCM_PRIORITIES`.

@NeoElit, what do you think? Since you also seem to use GCM and want this gem to support more of its features.

aried3r:
Implemented in #243.

PR #234, opened 2016-01-20
Fixed issue where setting the mdm parameter broke to_binary

troya2:
If it's an MDM push, no 'aps' key is present in the json, so the priority_for_frame method failed and prevented the push.  I added a spec that fails without this fix.


aried3r:
Thanks for contributing! Do you have a link to the developer docs where this is documented? I've, so far, not worked with MDM pushes.

troya2:
It's documented in a document only available to enterprise developer members.  I actually discovered the solution imperially by watching the iPhone console logs while implementing the MDM service.  The doc is titled "MDM Protocol Reference" from Apple.

-Troy

aried3r:
I don't have access to this document but from a small research, this checks out. Thanks!

PR #238, opened 2016-01-22
Add support for windows raw notifications (in json form)

mseppae:
Added some helpful information to get going with the Windows notifications.

### DIFF
--- a/lib/rpush/client/active_record/wns/raw_notification.rb
+++ b/lib/rpush/client/active_record/wns/raw_notification.rb
@@ -1,0 +1,13 @@
+module Rpush
+  module Client
+    module ActiveRecord
+      module Wns
+        class RawNotification < Rpush::Client::ActiveRecord::Notification
+          validates_with Rpush::Client::ActiveModel::PayloadDataSizeValidator,
+                         limit: 5120
+          include Rpush::Client::ActiveModel::Wns::Notification
+        end
+      end
+    end
+  end
+end
--- a/README.md
+++ b/README.md
@@ -114,7 +114,9 @@ n.save!
 
 For more documentation on [ADM](https://developer.amazon.com/sdk/adm.html).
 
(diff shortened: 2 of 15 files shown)


mseppae:
Oh man travis, haha.

aried3r:
Yeah, there's something weird going on with Travis wrt to Ruby 2.3.0. Rebuilding it fixes it.

mseppae:
Windows raw notification support finished. No more commits coming.

aried3r:
Great, thanks! I sadly cannot judge if this works since I've no way to test this myself, but it looks good, judging from the documentation.

Do you think you could squash your commits a bit?

aried3r:
It seems a payload cannot exceed 5KB, could you add a validation and specs for this? I think the other specs have it somewhere.

See https://msdn.microsoft.com/en-us/library/windows/apps/hh761463.aspx#recommendations

mseppae:
Added the validation and squashed the commit to a single commit.
This is the code we currently use for our windows phone notifications, so I can confirm that it works.

lib/rpush/client/active_record/wns/raw_notification.rb
```diff
@@ -0,0 +1,13 @@
+module Rpush
+  module Client
+    module ActiveRecord
+      module Wns
+        class RawNotification < Rpush::Client::ActiveRecord::Notification
+          validates_with Rpush::Client::ActiveModel::PayloadDataSizeValidator,
+                         limit: 5000
```
aried3r: Do you think they mean 5KB with base 10 (5 * 10^3 = 5000) and not 5KiB as in 5 * 2^10 = 5120 byte?

mseppae:
The payload limit is 2 based not 10, makes sense. Thanks for catching it.
Made sure by manually sending one byte too long payload.

PR #247, opened 2016-02-08
Various improvements for WNS notifications

wouterh:
Add support for:
* WNS badge notifications
* The launch argument of WNS toast notifications
* Sound in WNS toast notifications

### DIFF
--- a/lib/rpush/daemon/wns/toast_request.rb
+++ b/lib/rpush/daemon/wns/toast_request.rb
@@ -22,16 +22,23 @@ module Rpush
         def initialize(notification)
           @title = notification.data['title'] || ''
           @body = notification.data['body'] || ''
+          @launch = notification.data['launch']
+          @sound = notification.sound unless notification.sound.eql?("default".freeze)
         end
 
         def to_xml
-          "<toast>
+          launch_string = "" unless @launch
+          launch_string = " launch='#{CleanParamString.clean(@launch)}'" if @launch
+          audio_string = "" unless @sound
+          audio_string = "<audio src='#{CleanParamString.clean(@sound)}'/>" if @sound
+          "<toast#{launch_string}>
             <visual version='1' lang='en-US'>
               <binding template='ToastText02'>
                 <text id='1'>#{CleanParamString.clean(@title)}</text>
                 <text id='2'>#{CleanParamString.clean(@body)}</text>
               </binding>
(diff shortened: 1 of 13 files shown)


aried3r:
What do you think about leaving `validates :data, presence: true` in the base class but with a conditional, like `validates :data, presence: true, unless: skip_validation?`, where `skip_validation?` returns false by default? That way, further notifications (it seems to me, that all but the Badge Notification need the validation) would not need to remember to have `validates :data, presence: true` but those that don't need it have to intentionally disable it.

wouterh:
I will implement something like that tomorrow. Thanks for the feedback.

Op ma 8 feb. 2016 20:35 schreef Anton Rieder <notifications@github.com>:

wouterh:
@aried3r I have implemented your suggestion. I have called the check skip_data_validation? as it causes only the validation of :data to be skipped. The uri etc are still validated.

lib/rpush/daemon/wns/toast_request.rb
```diff
@@ -22,16 +22,23 @@ class ToastRequestPayload
...
           @title = notification.data['title'] || ''
           @body = notification.data['body'] || ''
+          @launch = notification.data['launch']
+          @sound = notification.sound unless notification.sound.eql?("default".freeze)
         end
 
         def to_xml
-          "<toast>
+          launch_string = "" unless @launch
+          launch_string = " launch='#{CleanParamString.clean(@launch)}'" if @launch
```
aried3r: Do you need the extra space before `launch`?

aried3r:
Looking good, just waiting for your answer regarding that extra space.

Also, could you squash the commits into one, once done? :)

lib/rpush/daemon/wns/toast_request.rb
wouterh: Yes, it is required. It is used a couple of lines below in:
```
"<toast#{launch_string}>
```

which is meant to produce either:
```xml
<toast>
```
or
```xml
<toast launch='...'>
```

aried3r: Ah, right. I missed that. Thanks for clarification!

aried3r:
Ah, one thing I forgot, would you mind updating the README accordingly to your changes? Would be much appreciated!

PR #266, opened 2016-03-22
Precise after init commit msg

azranel:
Just small adjustment to after init message. It's more clear I think. :smile:


aried3r:
Thanks!

PR #272, opened 2016-05-10
Basic Ionic Push service support.

Ynnni:
http://docs.ionic.io/docs/push-overview

### DIFF
--- a/lib/rpush/daemon/dispatcher_loop.rb
+++ b/lib/rpush/daemon/dispatcher_loop.rb
@@ -25,6 +25,7 @@ module Rpush
           loop do
             payload = @queue.pop
             if stop_payload?(payload)
+              p 2
               break if should_stop?(payload)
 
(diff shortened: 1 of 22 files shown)


lib/rpush/daemon/dispatcher_loop.rb
```diff
@@ -25,6 +25,7 @@ def start
           loop do
             payload = @queue.pop
             if stop_payload?(payload)
+              p 2
```
aried3r: Leftover?

aried3r:
Hey, thanks for the contribution. Would you mind adding some specs?

lib/rpush/daemon/dispatcher_loop.rb
Ynnni: Hey, thanks. I will remove this.

Ynnni:
Okay, i think this would easy.

PR #286, opened 2016-07-01
ERROR: "command not found: rpush" FIX: Bundle exec prefix

aymericbouzy:
Ran into "command not found: rpush" error, and it took me a quarter of hour to figure out why. Thought I'd save some time to people using a different ruby version like me and running into the same problem.
Feel free to reject the PR if I'm just an ignorant newbie who should know this stuff already.


aried3r:
LGTM. I think newcomers will appreciate it. :)

PR #296, opened 2016-08-30
Support APNS mutable-content

tdtran:
See Apple dev documentation https://developer.apple.com/reference/usernotifications/unnotificationserviceextension

I added a new AR migration `Rpush280Updates` assuming the next `rpush` version will be `2.8.0`.

PR #302, opened 2016-10-04
Prepare 2.8.0 release

tdtran:
As suggested by @aried3r (https://github.com/rpush/rpush/pull/296#issuecomment-249533569) here's my draft PR for rpush 2.8.0 release

/cc @ileitch


aried3r:
Travis had tests disabled. I re-enabled them, but there is no way to trigger a rebuild for this PR. At least I don't see one. We could close and re-open this PR.

amaierhofer:
I reverted this PR because it was causing build errors on travis, specifically redis builds. Do you have a good reason to upgrade the redis dependency?

tdtran:
Not really. I only noticed the deps in `Gemfile.lock` are quite old and thought it may be a good idea to update them.

Gemfile.lock
```diff
@@ -123,95 +125,99 @@ GEM
...
+    railties (4.2.7.1)
+      actionpack (= 4.2.7.1)
+      activesupport (= 4.2.7.1)
       rake (>= 0.8.7)
       thor (>= 0.18.1, < 2.0)
     rainbow (2.1.0)
-    rake (10.5.0)
-    redis (3.2.2)
+    rake (11.3.0)
+    redis (3.3.1)
```
amaierhofer: Can you simple change that line back to 3.2.2 and we have another go? I think that might work

PR #303, opened 2016-10-05
Switch GCM with FCM server endpoint

aried3r:
I purposedly left all comments, classes, debug output etc untouched,
because other than the server endpoint, users of the library still use
GCM and not FCM.

Renaming everything would break the API and thus require a version 3.

For reference, see:
https://developers.google.com/cloud-messaging/android/android-migrate-fcm#update_server_endpoints

### DIFF
--- a/lib/rpush/daemon/gcm/delivery.rb
+++ b/lib/rpush/daemon/gcm/delivery.rb
@@ -1,12 +1,12 @@
 module Rpush
   module Daemon
     module Gcm
-      # http://developer.android.com/guide/google/gcm/gcm.html#response
+      # https://firebase.google.com/docs/cloud-messaging/server
       class Delivery < Rpush::Daemon::Delivery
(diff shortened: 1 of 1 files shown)


aried3r:
The tests with Rubinius seem to fail because of an error with Travis-CI itself: https://github.com/travis-ci/travis-ci/issues/5294

aried3r:
Friendly ping @amaierhofer, would you mind taking a look? :)

amaierhofer:
Merging in good faith that you have verified it does send via the new endpoint :) 

I believe we should ignore rbx on travis for now, I believe adaption rates are low

jaz5hq:
so I am wondering, if i use the documented way for implementing gcm to implement fcm. It will work because of the change of the endpoint?

aried3r:
@jaz5hq Would have worked even without the endpoint change ;)

PR #310, opened 2016-10-24
Loosen net-http-persistent dependency

aried3r:


aried3r:
Shouldn't be too hard, but I currently don't have the time to pursue this. Closing.

PR #311, opened 2016-11-04
Changed migration to remove default sound

amaierhofer:
According to official documentation the sound key must not be included when sending silent push.

See Silent Push Section in Official Docs

https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html#//apple_ref/doc/uid/TP40008194-CH10-SW1


aried3r:
If we do this in 2.8.0, wouldn't this violate http://semver.org?

amaierhofer:
You are right. We should hold back until we can change the major version.

PR #318, opened 2016-11-24
Run ApnsFeedback only on Apns apps.

robertasg:
Currently, ApnsFeedback tries running on every added application, including GCM for example. This results in a crash due to incompatible interfaces.


amaierhofer:
I do not think this is necessary. 

`Rpush::Apns::App.all` already ensures that only Apns apps are considered when starting the Feedback Receiver.

robertasg:
Not on Redis. On Redis every App is on the same namespace defined in Rpush::Client::Redis::App. 
Perhaps this can be solved with a different approach, eg. using different namespaces throughout the different apps to be more consistent.

amaierhofer:
Ahh I see, it's all legit :) 

I would appreciate if you could add a comment explaining that this is necessary when using Redis and add a line to the changelog file.

amaierhofer:
Great! Many thanks!

PR #326, opened 2017-01-09
Update APNS App creation example's environment name.

chris:
* The APNS environment names in the latest code are "development" and "production" (as seen here: https://github.com/rpush/rpush/blob/master/lib/rpush/daemon/dispatcher/apns_http2.rb#L8), so using "sandbox" will fail with a Bad URI error. Changes the README's example to use "development" so that it'll work with the latest code.


amaierhofer:
Thanks for your contribution

PR #328, opened 2017-01-16
Apns (http2) performance fix - use batches

Nattfodd:
Performance is increased because of:
- now we prepare batch of requests and send them at once, instead of sending each Notification separately.
- now we use multi-threading to process requests concurrently 

Overall, on a 5000 notifications with a batch size of 2024, sending time reduced from ~1500 seconds to 10 seconds.

### DIFF
--- a/lib/rpush/daemon/apns2/delivery.rb
+++ b/lib/rpush/daemon/apns2/delivery.rb
@@ -37,7 +37,7 @@ module Rpush
         ######################################################################
 
         def prepare_async_post(notification)
-          response = OpenStruct.new
+          response = {}
 
           request = build_request(notification)
           http_request = @client.prepare_request(:post, request[:path],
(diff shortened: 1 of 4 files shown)


lib/rpush/daemon/apns2/delivery.rb
```diff
@@ -8,124 +8,132 @@ module Apns2
...
-
-        def service_unavailable
-          mark_retryable(@notification, Time.now + 10.seconds)
-          # Logs should go last as soon as we need to initialize
-          # retry time to display it in log
-          failed_message_to_log
-          retry_message_to_log
-        end
+        def prepare_async_post(notification)
+          response = OpenStruct.new
```
amaierhofer: Performance wise I think it would be better use a Struct or Hash here as OpenStruct uses method_missing internally, see http://ruby-doc.org/stdlib-1.9.3/libdoc/ostruct/rdoc/OpenStruct.html#class-OpenStruct-label-Implementation-3A

Nattfodd:
The numbers above are for Redis storage.

amaierhofer:
Many thanks for speeding up things :)

PR #332, opened 2017-01-26
[APNS2] Notification doesn't change its status fix

Nattfodd:
Fixed the bug that caused notifications to be saved before their status has been changed to some (delivered/failed).

```ruby
def prepare_async_post(notification)
  # ... prepare single request to APNS

  # [Wrong] 
  # This line calls `complete` as soon as the last notification
  # has been marked as processed
  # But on this stage we didn't send any request yet
  # (send all at once with `@client.join`)
  # So notifications have been marked as processed & have been saved without any status
  # and we never return to it anymore
  @batch.notification_processed
end
```
```ruby
def perform
  # ...prepare requests for all notifications in the batch...
  # ... send them at once ...
  @client.join
  # [Wrong] Every notification status updates independently,
  # since each has it's own response
  @batch.mark_all_delivered
end

### DIFF
--- a/lib/rpush/daemon/apns2/delivery.rb
+++ b/lib/rpush/daemon/apns2/delivery.rb
@@ -21,8 +21,6 @@ module Rpush
 
           # Send all preprocessed requests at once
           @client.join
-
-          @batch.mark_all_delivered
         rescue SocketError => error
           mark_batch_retryable(Time.now + 10.seconds, error)
           raise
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Nattfodd** [comment]
@amaierhofer @aried3r would you please review this one? It might be my fault the description to this PR is confusing a bit. But in fact it's just kinda typo fix.

File: lib/rpush/daemon/apns2/delivery.rb [hunk 1]
```diff
@@ -21,8 +21,6 @@ def perform
 
           # Send all preprocessed requests at once
           @client.join
-
-          @batch.mark_all_delivered
```
aried3r: When do they get marked as delivered once this line is removed?

aried3r: Nevermind, you explain it in your post.

PR #351, opened 2017-04-12
[APNS2] Retry delivery on networking errors

Nattfodd:
Net/Http2 gem processes delivery in a separate thread.
When it dies, it crashes the whole Rpush process,
which isn't intended.

There's a way to callback network errors.


amaierhofer:
Thanks for that!

Issue #260, opened 2016-02-26
Android Rpush Notification doesnt get called when app is in background after set content_available


boyfunky:
So this is my code to get my notification event to be fired when the app is in background on the client side but it doesnt work. 

```
n = Rpush::Gcm::Notification.new
n.app = Rpush::Gcm::App.find_by_name("partners")
n.registration_ids = [@user_auth_token.device_token]
n.data = { state: "case_status_single", count: badge_unread.length + 1, notification_id: n.id, content_available: '1'}
n.content_available = true
n.notification = { 
    body: @loan_application.name + " Loan Status Changed to: " + AgentAssignmentStatus.find(p.agent_assignment_status_id).assignment_status,
    title: Partners',
    tag: true
 }
 n.save!
 Rpush.push
```

I thought with the content_available parameter allows the onNotification event from the mobile app side to be triggered even if the app is in background. 

Any thoughts on this?

Thanks

aried3r:
`content_available` is only for iOS currently. On Android, the data message wakes up the device anyway. There is however the `priority` which can be either `normal` or `high`. `high` should wake up even devices with Doze enabled. Note: You usually do not want to send all messages as high as it will always wake up the device and would abuse the service.

For more about the priority, read more here:
https://developers.google.com/cloud-messaging/concept-options#setting-the-priority-of-a-message

Also the quote of Google Engineer, Dianne Hackborn:
> Abuse of high priority messages have a special difference from other things like notifications: they must go through Google servers, so Google can monitor and modify what is being sent to devices.  If apps abuse these for other things besides their intended use, we will be able to stop that abuse without touching any software on the device.  (Also "abuse" here is much less subjective than for notifications, where there is a large gray area of things some users care about and some don't.  For high priority messages, if it isn't something that is time critical to go to the user immediately, it is not appropriate.)

https://plus.google.com/105051985738280261832/posts/4x9tJoFWNof

boyfunky:
@aried3r yes i understand that android should wake up the device but it doesnt. It only catches the onNotification event when the app is in the foreground and doesnt wake it when it is in the background. So how can i handle this?

aried3r:
I don't think this is a problem with rpush but your Android app. You could try to verify by manually sending the same payload via Postman.
We use the same kind of payload for our app at work, so I don't think it's that.

fertobar:
hello, I notice I set  priority **high** for GCM, but on Payload I see **priority=>10** , that is ok for GCM?

[...]

PR #353, opened 2017-04-24
fix GCM priority

fertobar:


aried3r:
Closing, see https://github.com/rpush/rpush/issues/260#issuecomment-367708696

PR #354, opened 2017-04-26
[APNS2] Set default timeout for async http2 requests

nattfodd:
**Note:** _that's not a perfect solution, but to do it properly net-http2 gem is required to be updated. There's the [issue](https://github.com/ostinelli/net-http2/issues/19) opened, but it has no estimates._ 

We found another issue with delivering apns notifications via http2. Since async requests have no built-in timeout, if any of notifications stucks, the entire batch waits for it (`@client.join`), and it may take up to few hours with a SockerError at the end.

To avoid that, we introduce timeout for the batch. Notifications will be still delivered in parallel using HTTP2 async streams, but since rpush saves notifications statuses once after delivery of entire batch is completed, we have to wait and retry some notifications that have been failed by timeout and allow further notifications to be processed with a minimal delay.

### DIFF
--- a/lib/rpush/daemon/apns2/delivery.rb
+++ b/lib/rpush/daemon/apns2/delivery.rb
@@ -7,11 +7,14 @@ module Rpush
 
       class Delivery < Rpush::Daemon::Delivery
         RETRYABLE_CODES = [ 429, 500, 503 ]
+        DEFAULT_TIMEOUT = 60
 
         def initialize(app, http2_client, batch)
           @app = app
           @client = http2_client
           @batch = batch
+
+          @processed_statuses = {}
         end
 
         def perform
@@ -21,11 +24,20 @@ module Rpush
             prepare_async_post(notification)
           end
 
-          # Send all preprocessed requests at once
-          @client.join
+          # Async http2 requests don't have a timeout,
+          # so it's important to implement it on a higher level
+          Timeout.timeout(DEFAULT_TIMEOUT) do
+            @client.join
+          end
         rescue Errno::ECONNREFUSED, SocketError => error
           mark_batch_retryable(Time.now + 10.seconds, error)
           raise
+        rescue Timeout::Error => error
+          @batch.each_notification do |notification|
+            next if @processed_statuses[notification.id]
+
+            service_unavailable(notification, error)
+          end
         rescue StandardError => error
           mark_batch_failed(error)
           raise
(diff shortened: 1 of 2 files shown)


aried3r:
Hey @Nattfodd, I'm not using APNS2 myself, but have you successfully used this code in production? If so, I'd merge it so others can profit from it!

Nattfodd:
@aried3r we don't use this fix on production yet, we just tested it with our QA team. That's not a perfect solution though (issues with net-http2 gem), but it works.

ostinelli:
> issues with net-http2 gem

What issues? :)

aried3r:
Hey @Nattfodd, just wanted to check in again if you can confirm this patch is working as intended for you. Maybe we can resolve this for rpush and should there be a bug in net-http2, @ostinelli seems to be interested to resolve this as well. :)

Nattfodd:
@aried3r we don't use this patch in production yet. As far as I can say from QA results, it works, we tested delivery of a pretty big amount of push notifications (like ~300 per second). I don't know when we will be finally release this one, so no estimates. I'm not really confident in HTTP2, so I can't suggest a better solution for the moment. At least this one prevents deliveries being stuck for hours in case of any networking issues. In theory, this PR may cause same notification to be delivered more than once, since network may "unstuck" and deliver notification, but we already marked it as failed after 60 seconds timeout.

It's not a bug in net-http2, it's a missing feature I would say.

ostinelli:
> @ostinelli seems to be interested to resolve this as well. :)

I still don't see what issues are supposedly with NetHttp2 which would need something to be solved. If anything is needed there, please open an issue on the gem's repo. :)

aried3r:
@ostinelli Thanks for offering your help! Could I ask you to take a look at https://github.com/rpush/rpush/pull/412? It seems this error handler is definitely something you recommend, but for some reason makes all the tests fail.

@armahmoudi, you seem interested in this as well. But again, I don't use this feature myself so I can't speak for the code. Please also see #412 which I mentioned above.

Any help is greatly appreciated!

jhottenstein:
@nattfodd, we are running into this issue in production so I've been looking into solutions.  I like your idea but I share the intuition of @ostinelli that we should avoid using the Timeout class.  

I'm trying out a IO.select timeout in the net-http2 socket loop and if that doesn't work, I had the idea to add a simple elapsed time timer inside the join method.

Do you have a preference?

PR #362, opened 2017-06-16
Reconnect on AR ConnectionNotEstablished

daniel-illi:
Sometimes we get `ActiveRecord::ConnectionNotEstablished` errors on our production system.
If this happens, Rpush does not try to reconnect to the database, and the currently processed notifications batch fails.

This pull request adds `ActiveRecord::ConnectionNotEstablished` to the list of reconnectable errors.

### DIFF
--- a/lib/rpush/daemon/store/active_record/reconnectable.rb
+++ b/lib/rpush/daemon/store/active_record/reconnectable.rb
@@ -22,7 +22,8 @@ module Rpush
         module Reconnectable
           ADAPTER_ERRORS = [::ActiveRecord::StatementInvalid, PGError, PG::Error,
                             Mysql::Error, Mysql2::Error, ::ActiveRecord::JDBCError,
-                            SQLite3::Exception, ::ActiveRecord::ConnectionTimeoutError]
+                            SQLite3::Exception, ::ActiveRecord::ConnectionTimeoutError,
+                            ::ActiveRecord::ConnectionNotEstablished]
 
           def with_database_reconnect_and_retry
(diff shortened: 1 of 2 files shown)


aried3r:
Great, thank you! Could you try to write a spec for it? I guess https://github.com/rpush/rpush/blob/master/spec/unit/daemon/store/active_record/reconnectable_spec.rb would be the place.

mfbmina:
@daniel-illi will you do it or can I do?

daniel-illi:
@aried3r as requested I have added a spec, but I'm not sure about its value as it only tests if the specified errors are handled by `Reconnectable`. It will only ever fail if someone removes an error class from the hardcoded list in `Reconnectable`.

@mfbmina would you take a different approach to test this change?

aried3r:
> but I'm not sure about its value

It has questionable value, yes. I would love to have proper tests for this, but I'm not sure how. Let's see if @mfbmina has an idea, otherwise I'm gonna merge this as-is.

mfbmina:
It seems ok for me! 👍

mfbmina:
If you add create a test case that raises an error (which is not hard coded), then you can test that is not reconnecting.. Can this help?

daniel-illi:
@mfbmina good point, if I test that it is reconnecting on the defined errors, I probably should test as well that it is not reconnecting on a different error.

daniel-illi:
I added the spec for non-reconnectable errors.
What do you think @aried3r, is it ready to merge?

aried3r:
LGTM

PR #366, opened 2017-06-26
Remove legacy Rails and Rubies

aried3r:
I think even though this gem is kinda-sorta unmaintained (as in, there are no new releases), I think it's time we clean up some stuff.

Things I want to try to do for rpush "4.0" (or 3.0, if we want to add breaking changes there, but I think a lot of people are already using the `master` branch):
- Only support Rails 4.2+
- Only support Ruby 2.2+
- Drop JRuby support (debatable, it's just that the test always fail, so I have little confidence in it)
- Add Appraisals or another way of testing different Rails versions

I'm open for discussion!

Fixes #364

PR #367, opened 2017-06-28
Upgrade modis dep on Gemfile.lock

Tonkpils:
I was wondering if I should be updating the rails 3 Gemfile.lock /cc @aried3r


Tonkpils:
Seems like the failing specs are also failing on master

Tonkpils:
@aried3r that second commit is what was failing on the specs. Modis 1.4.2 changed the namespaces where records were being stored in Redis to account for STI namespaces but this gem was not. I am wondering/thinking that modis should probably provide a method that exposes that namespace directly rather than duplicating the logic here...

See https://github.com/ileitch/modis/blob/v1.4.2/lib/modis/persistence.rb#L244

aried3r:
> I was wondering if I should be updating the rails 3 Gemfile.lock

Hmm, personally I wouldn't, because I really wanna drop Rails 3 support.

> I am wondering/thinking that modis should probably provide a method that exposes that namespace directly rather than duplicating the logic here...

If you're up for it, sure! But until that gets released, I guess we have to duplicate logic.

Btw, can we also use modis 2.1.0 in the current rpush version (`master`) or is something missing?

Tonkpils:
> Btw, can we also use modis 2.1.0 in the current rpush version (master) or is something missing?

Notable changes are
- Dropped EOL ruby support
- Removed yaml encoding/decoding and use msgpack instead (this was part of 1.x but was hidden behind a flag)

Currently, `rpush-redis` master has modis locked to version 2.x which would mean if someone installs rpush-redis from master they'd be using modis. 

Since `modis` is a dep of `rpush-redis` we'd have to bump that first. The only issue I could foresee with the upgrade is people that have records with YAML will not work.

aried3r:
I see. Is there a migration guide for YAML -> msgpack? Or will that resolve itself over time? I'm not sure how you use `msgpack` there.

Going forward, I'd love to see `rpush-redis` having a new major release and use that in our specs here, although I don't use it. Do you think @ileitch would be up for it or do you have access to the repo?

Tonkpils:
I have access to the repo but not to release a new version as I'm not part of the project in Rubygems. @ileitch if you'd like I could schedule a new release for it!

aried3r:
I should have release permissions for modis now, I'd love to see a new version released, think we can add a changelog fro 1.x to 2.x and possible caveats/migration guide?

I'm now working on #366 again and hopefully prep a 3.0.0.rc1 release of rpush soon :)

aried3r:
Hey @Tonkpils, I merged #378, it's a different approach than yours. Mind taking a look? :)

Issue #372, opened 2017-07-25
Mismatched pending namespace between Store::Redis and Redis::Notification


czhc:
```
gem 'rpush', '~> 2.7.0'
gem 'rpush-redis', `~> 0.4.1`
```

We are running `rpush` as a daemon, using redis as a store.

All instances of `Rpush::...::Notification.new` return `true` on `.save!`
But no notifications are added to queue.

What we noticed was there was no entry for `rpush:notifications:pending`, but instead the notifications were added to a key `rpush:notifications:notification:pending`.

```
127.0.0.1:6379> keys rpush:notifications:*
...
2) "rpush:notifications:notification:all"
6) "rpush:notifications:notification:pending"
...
```

Manually adding the notification into a key `rpush:notifications:pending` successfully enqueues the notification for delivery.

When we run on console:

```
Rpush::Client::Redis::Notification.absolute_pending_namespace
=> "rpush:notifications:pending"
```

Not sure what is going wrong here.

Nattfodd:
You're probably using new Modis gem version ( >= 1.4.2). 1.4.2 version of this gem introduced some fixes related to STI Childs logic, including namespaces changes (https://github.com/ileitch/modis/commit/3d7f0b3a1fcfad98587be8a86d5d61b7a4cb0d19).

One solution is to lock Modis on 1.4.1 version. Another one, which we actually did - to monkeypatch Rpush::Notification class to explicitly set namespace to what it was before:

```ruby
require 'rpush/client/redis'

classes = Rpush::Client::Redis::Notification.subclasses + [Rpush::Client::Redis::Notification]
classes.each do |notification_class|
  notification_class.class_eval do
    self.namespace = 'notifications'
  end
end
```

czhc:
Thanks @Nattfodd, wasn't looking into `modis`. I'll leave the issue open if there isn't an existing patch for `rpush` or `rpush-redis` already.

PR #378, opened 2017-09-12
Update Modis to 1.4.2

milgner:
As documented in [#372](https://github.com/rpush/rpush/issues/372) the notifications namespace of the subclasses had changed, preventing them from being delivered.


aried3r:
Thanks!

Tonkpils:
This looks good, I don't believe this will be an issue with rpush since the namespace is only used for the notification class.

Issue #382, opened 2017-10-07
make booleans true or false ... not nil


grosser:
mutable_content, content_available and alert_is_json can be nil/true/false which seems dangerous ...
would you accept a PR that changes them via `change_column_null false` ?

FYI found via:

```Ruby
  it "does not have 3-state booleans (nil/false/true)" do
    bad = File.read("db/schema.rb").scan(/\st\.boolean.*/).reject { |l| l .include?(" null: false") }
    assert bad.empty?, "Boolean columns missing a default or null: false\n#{bad.join("\n")}"
  end
```

aried3r:
Sure, I don't see why not. Would you change the migration templates or add a `3_0_1_updates` migration? Asking because I'm not sure if this should be seen as a backwards incompatible breaking change which would demand a 4.0 according to semver. I don't know if people rely on it being `nil` at any point.

In theory it should be an implementation detail which should not break stuff.

PS: That's a neat test, got some more of those sanity checks online?

grosser:
Was thinking about adding a new migration to make it backwards compatible ... doubt it will break anything.

other tests are here https://github.com/zendesk/samson/blob/master/test/integration/cleanliness_test.rb ... few reusable ... most very project specific :)

aried3r:
Fixed in #384

PR #384, opened 2017-10-21
reduce booleans to true/false, do not allow nil

grosser:
solves #382

### DIFF
--- a/lib/generators/templates/rpush_3_0_1_updates.rb
+++ b/lib/generators/templates/rpush_3_0_1_updates.rb
@@ -1,0 +1,13 @@
+class Rpush301Updates < ActiveRecord::VERSION::MAJOR >= 5 ? ActiveRecord::Migration[5.0] : ActiveRecord::Migration
+  def self.up
+    change_column_null :rpush_notifications, :mutable_content, false
+    change_column_null :rpush_notifications, :content_available, false
+    change_column_null :rapns_notifications, :alert_is_json, false
+  end
+
+  def self.down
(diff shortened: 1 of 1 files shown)


grosser:
@aried3r

aried3r:
Thanks!

lib/generators/templates/rpush_3_0_1_updates.rb
```diff
@@ -0,0 +1,13 @@
+class Rpush301Updates < ActiveRecord::VERSION::MAJOR >= 5 ? ActiveRecord::Migration[5.0] : ActiveRecord::Migration
+  def self.up
+    change_column_null :rpush_notifications, :mutable_content, false
+    change_column_null :rpush_notifications, :content_available, false
+    change_column_null :rapns_notifications, :alert_is_json, false
```
redross: I'm most likely doing something wrong (late night coding :smile: ), but this part fails for me locally (the table `rapns_notifications` doesn't exist). Shouldn't it be:
```
change_column_null :rpush_notifications, :alert_is_json, false
```

grosser: yes!

... table got renamed :(

https://github.com/rpush/rpush/blob/2fa09cc352eae4abff8be7e0eec6324786df6e7f/lib/generators/templates/rename_rapns_to_rpush.rb

grosser: https://github.com/rpush/rpush/pull/396

PR #387, opened 2017-11-07
better error message for error 8

grosser:
according to https://github.com/node-apn/node-apn/issues/135 it's an invalid device token ... since we deal with lots of tokens that should make debugging faster ...

@aried3r


aried3r:
Thanks!

PR #390, opened 2017-11-30
In Rails 5.2 passing default to #index_name_exists? was removed.

biow0lf:
Changes:

* Added rails 5.2 and rails master to travis build matrix
* Wrap `#index_name_exists?` with rails version check

Rails 5.2 CLIENT=redis test suit fails. Need some more love. ❤️ I will add more commits for this later in this PR, or open another PR.


biow0lf:
Looks like I break rails 4.2. Fixing.

aried3r:
Thanks @biow0lf! Looks good so far. Do you think you can get 5.2 to work with redis?

/cc @Tonkpils

aried3r:
Hi, could you rebase this branch? There's a few fixes in `master` that should help with some of the CI failures. Then we can have another look :)

biow0lf:
Sure, I will rebase it. Also, I want update update modis gem to work with rails 5.2. But, I need some time for this.

Tonkpils:
Let me know if there's anything I can do to move this effort forward. I'd be happy to 👀 or pair on this work 😄

aried3r:
Hi, I rebased this branch and now only the tests with Rails 5.2 with fail. Not sure yet what's causing them.

aried3r:
I think this PR is fine for AR as a backend, since the tests work. But it seems the redis errors are unrelated to that, since they don't use the migrations.

Might be worth merging this and working on the redis issue separately. What do you think?

biow0lf:
> Might be worth merging this and working on the redis issue separately. What do you think?

redis backend broken due broken support of rails 5.2 in gem modis.

soulfly:
We face the same issue with modis support in rails 5.2
https://github.com/ileitch/modis/issues/13

Does anybody have a clue when this can be done?

We can't update our rails app from 4.2 to 5.2 because of this issue with modis in rpush

aried3r:
@soulfly, no idea, sadly. I don't use modis at all. Are there actively maintained alternatives out there we could switch there in case modis does not receive an update?

I'm thinking about accepting this PR since it seems to work with ActiveRecord and writing a notice in the README that for now, it doesn't work with MongoDB.

soulfly:
@aried3r 

>  it doesn't work with MongoDB.

You mean Redis

not sure about alternatives, Modis is something that integrated with ActiveRecord API so you do not need to learn any new API with it

@OL-Mark  do you have any thoughts?

aried3r:
> You mean Redis

Yes, you're right, I've corrected it in my comment, thanks!

aried3r:
Sorry for taking so long. I'll have another look and add a notice to the README that Redis support currently doesn't exist for Rails 5.2 with Modis.

Hope to get it out soon, thanks for the PR @biow0lf and everyone else for their help!

Issue #106, opened 2014-11-18
Difference in environments


jayroh:
Hey there. In the README it mentions that, when creating a new "app", you set the environment for the app:

``` ruby
app = Rpush::Apns::App.new
app.name = "ios_app"
app.certificate = File.read("/path/to/sandbox.pem")
app.environment = "sandbox" # <-- HERE
app.password = "certificate password"
app.connections = 1
app.save!
```

When looking through the validations for the `Rpush::Apns::App` model I see that it validates one of three values - `development`, `production` or `sandbox`.  (in `lib/rpush/client/active_model/apns/app.rb`).

I've been looking for documentation, both here and on Apple's site, that might give some clarification as to what those differences are, but I don't see anything concrete. Is sandbox the same as development? Should I use "production" for an in-house distribution? Is there anywhere that might have more information on this?

I'd be more than happy to wrangle some of this up and submit a PR for the README if it helps.

PR #392, opened 2017-12-03
Add sandbox URL to ApnsHttp2 dispatcher

brianlittmann:
Was getting `URI::InvalidURIError, bad URI(is not URI?)` when trying to send Apns2 notifications with `app.environment` set to `sandbox`.

### DIFF
--- a/lib/rpush/daemon/dispatcher/apns_http2.rb
+++ b/lib/rpush/daemon/dispatcher/apns_http2.rb
@@ -5,7 +5,8 @@ module Rpush
 
         URLS = {
           production: 'https://api.push.apple.com:443',
-          development: 'https://api.development.push.apple.com:443'
+          development: 'https://api.development.push.apple.com:443',
+          sandbox: 'https://api.development.push.apple.com:443'
         }
 
         DEFAULT_TIMEOUT = 60


aried3r:
Hey and thanks for your PR.

I don't use APNS myself, but it seems the endpoints are just development and production. https://developer.apple.com/library/content/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CommunicatingwithAPNs.html

According to https://github.com/node-apn/node-apn/pull/528 this got changed at some point, sandbox being removed.

Do you think this change is necessary considering the docs?

brianlittmann:
> Do you think this change is necessary considering the docs?

I would say so, at least in considering consistency within the codebase. See this (albeit old) [comment](https://github.com/rpush/rpush/issues/106#issuecomment-66178362) and the [current validation for the APNs app model](https://github.com/rpush/rpush/blob/c938417fb2e2ac9b341e4c6c4c0ff38099103d9f/lib/rpush/client/active_model/apns/app.rb#L8).

gregblake:
Our team recently encountered the same issue. We were getting `URI::InvalidURIError, bad URI(is not URI?)` when trying to send Apns2 notifications with `app.environment` set to `sandbox`. In other words, notifications worked fine in our application, unless the `app.environment` was `sandbox`.

This PR was very helpful in figuring out the cause of this issue. I wanted to provide feedback on how we resolved it in our application. Hopefully this will help others that may run across the same thing. The short answer is that we only needed to make changes to how data was being stored.

As @aried3r mentioned, it seems that Apple changed the hostname of their APNS server. What was once `sandbox` is now `development`. *This is what caused the issue.* From what I can tell, Apple introduced this change with APNS2. 

As @brianlittmann said, the [validations require the environment to be one of the following](https://github.com/rpush/rpush/blob/c938417fb2e2ac9b341e4c6c4c0ff38099103d9f/lib/rpush/client/active_model/apns/app.rb#L8): `development`, `production`, or `sandbox`.

We were able to resolve the issue by taking the following steps:

1. Our Rails application has a model that contains our APNS certs. Some of the records in this table had an `environment` of `production`, and others had an environment of `sandbox`. We changed `sandbox` to `development`. We did not change the records that had an environment of `production`, because those apps were working fine, and were unaffected by Apples change to their host name. We only had to change the "sandbox" certs, so that they used Apple's updated hostname of "development".
2. Our Rails application has another model named `MobileDevice`. This model has a string attribute named `apns_environment`. Prior to this change, the iOS devices would set the `apns_environment` to either `sandbox` or `production`. We made changes to the iOS app so that it now sets the `apns_environment` to either `development` or `production`. In the cases that we were using the `sandbox` name before, now we are using the name `development`, so it matches the changes Apple made to their hostname.

Once we made the 2 changes above, the [`app.environment`](https://github.com/rpush/rpush/blob/master/lib/rpush/daemon/dispatcher/apns_http2.rb#L17) matched [one of these 2 urls](https://github.com/rpush/rpush/blob/master/lib/rpush/daemon/dispatcher/apns_http2.rb#L6-L9), and notifications worked for both `development` and `production` environments.

aried3r:
@gregblake, thanks for the long and elaborate answer! What do you suggest we do? Personally I'd rather document this properly somewhere and leave it as is. Or even better, document it and leave a deprecation warning, but I doubt people will see it if it runs as a daemon.

ileitch:
Merging this. There's no harm in a duplicate environment name, but there is harm in `sandbox` causing an error.

PR #396, opened 2017-12-29
use actual table name

grosser:
fixup for https://github.com/rpush/rpush/pull/384

@aried3r @redross


grosser:
... also means we are missing any test that actually runs these migrations :trollface:

redross:
The commit name could be a bit better and maybe worth creating a follow-up issue about "testing migrations".

aried3r:
Yeah, I haven't looked at the test suite yet, not sure when I'll find the time. But it's weird that migrations aren't run. Makes me worry a bit.

I'll merge and try to release a 3.0.2 soon.

PR #412, opened 2018-02-15
Add minimal error handler for Net::Http2 client reqeust

spacebat:
This is to prevent things like Errno::ECONNRESET from terminating the
dispatcher in daemon mode, which leaves the notifications stuck
forever in the processing state.

Perhaps more can be done here, whether to mark the notification for
retry or failure is perhaps something for the user to configure.

See https://github.com/ostinelli/net-http2/issues/23#issuecomment-361616710

### DIFF
--- a/lib/rpush/daemon/apns2/delivery.rb
+++ b/lib/rpush/daemon/apns2/delivery.rb
@@ -57,6 +57,10 @@ module Rpush
 
           http_request.on(:close) { handle_response(notification, response) }
 
+          http_request.on(:error) do |error|
+            log_error("Error sending notification #{notification.id}: #{error}")
+          end
+
           @client.call_async(http_request)
         end
 


spacebat:
Some background: we've been using rpush driven by sidekiq-cron for some time and it seems to be working OK. Sometimes we'd get a network error mostly from APNS such as connection reset by peer. Intending to have more timely notifications and to be able to scale more, switching our deployment to run rpush in daemon mode as a system service was tried. However, it seems when in daemon mode, connection reset by peer causes the dispatcher to crash, leaving some jobs stuck in the state where processing is true, and they are never retried.

aried3r:
Hey! Thanks for your PR, LGTM, although I don't use APNS2. The comment on net-http2 seems clear enough. /cc @nattfodd, if you could take a look as well? No obligations of course :)

Could you rebase this so we can try to get the CI tests to pass?

nattfodd:
Hey there! It seems like this is the same issue, I tried to address here: https://github.com/rpush/rpush/pull/354

If your @spacebat solution works (I can't check it anymore, unfortunately) - I like it more, since my PR involves with `Timeout` module, which is more like a hack. You may take a look at the PR I mentioned to implement retries.

spacebat:
@aried3r I have rebased onto master but travis is still failing, I'm fairly new to Ruby and travis and I've no idea why.
@nattfodd this approach won't detect timeouts, but I would like to resolve that too. I think Net::Http2 should offer such an option. Perhaps we should work on a patch for that project.

aried3r:
Could you maybe take another look at this PR? It seems to fail all versions of Rails and Ruby, maybe it's something obvious.

aried3r:
@goalaleo, no status update. I don't use APNS myself, so I can't really test this in a real environment. I'm happy for any patches, though! But as you can see, this PR seems to break each and any test run, so I don't feel comfortable merging it. :\

Maybe someone can look into it some more? I'll also try to add a note that this project needs APNS contributors or maintainers. I'm currently just keeping this project alive but heavily rely on community contributions. :)

Issue #113, opened 2014-12-22
APNS: RPush thinks everything is okay, notification gets marked as delivered, but nothing arrives on phone


maxhollmann:
I set up an app with production environment, took a production device token and sent a couple of notifications via Rpush. Everything seems to be okay, the notification gets marked as delivered, the log only shows "id sent to token". But nothing happens. Nothing arrives on the phone. Feedback is empty, too.

The exact same setup with the pushmeup gem works, the push arrives within a second of sending it out.

I have no clue, any ideas what this could be?

ileitch:
I've just released `2.3.0.rc1`, can you please give it a try?

ileitch:
@maxhollmann any update?

dc85:
I also noticed this problem, not sure when it started. But the strange thing is, I have a few different way of creating push notifications, one of these ways I can confirm still work, but the other ways seems to be failing.

I run a daemon to send out the push notifications once they're created.


Failing case
=======================
n = Rpush::Apns::Notification.new
n.app = Rpush::Apns::App.find_by_name("my_app_name")
n.device_token = push_token
n.alert = content.truncate(45)
n.badge = push_counter
n.sound = "default"
n.deliver_after = created_at if created_at != nil
n.data = {
              :cat => category,
              :act => action,
              :oid => org_id,
              :src => source,
              :sid => source_id,
              :sdr => sender,
              :rpt => recipient
}
n.save!

Success case
==================
n = Rpush::Apns::Notification.new
    n.app = Rpush::Apns::App.find_by_name("my_app_name")
    n.device_token = push_token
    n.alert = message.truncate(50)
    n.badge = push_counter
    n.sound = "default"
    n.data = {
      :cat => cat,
      :oid => oid,
      :sid => sid,
      :sdr => sdr,
      :chd => chd
    }
    n.save!

This is the only difference in my whole process as far as I can remember.

I updated to 2.3.1 and it is still failing for the fail case

ileitch:
So even in the failing case the notification gets marked as delivered?

dc85:
Yup,

Also, a different section of my code, which uses the same function to send out the push as the failed case, is succeeding,

Its mind boggling what is happening. everything gets marked as sent but nothing actually is seen on the phone for the failed case.

ileitch:
Does it fail if you use the same device token for both workflows?

dc85:
All my test was done under the same devices

[...]

PR #419, opened 2018-02-28
improving documentation to indicate that title or body should be present in FCM

rodriggochaves:
Hello! I recently started to use Rpush in my project. In the documentation for GCM/FCM, it's says

```ruby
# Optional notification payload. See the reference below for more keys you can use!
n.notification = { 
  body: 'great match!',
  title: 'Portugal vs. Denmark',
  icon: 'myicon'
}
```

Initialy, as I was trying to make a simple sample to my notifications work, I removed this field thinking all keys are optional. 
Then, I used `$ rpush push` to send my recently created notifications but none arrived. I checked using `rails c` and all `notification.delivered` are true. The device that I'm using to receive the notifications is iOS.

After a few time debugging, I tried to add 

```
n.notification = { 
  body: 'example',
}
```

 in my logic and suddenly started to work. This led me to believe that `body` or `title` should be present to a notification appear to the user in his device. At least in iOS.

There're a [issue](https://github.com/rpush/rpush/issues/113) that demonstrates the same behavior.

Then, I created this pull request in hope to future developers don't make the same mistake. I'm open to talk about the problem that I describe too!

Hope it helps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ n.registration_ids = ["..."]
 n.data = { message: "hi mom!" }
 n.priority = 'high'        # Optional, can be either 'normal' or 'high'
 n.content_available = true # Optional
-# Optional notification payload. See the reference below for more keys you can use!
+# You must define at least `body` or `title` fiels to your notification appear. See the reference below for more keys you can use!
 n.notification = { body: 'great match!',
                    title: 'Portugal vs. Denmark',
                    icon: 'myicon'


README.md
```diff
@@ -98,7 +98,7 @@ n.registration_ids = ["..."]
 n.data = { message: "hi mom!" }
 n.priority = 'high'        # Optional, can be either 'normal' or 'high'
 n.content_available = true # Optional
-# Optional notification payload. See the reference below for more keys you can use!
+# You must define at least `body` or `title` fiels to your notification appear. See the reference below for more keys you can use!
```
wladimirgramacho: I have a few improvements for this message:
```
# You must define at least `body` or `title` fields to make your notification appear. See the reference below for more keys you can use!
```

PR #423, opened 2018-03-05
Require 'rpush'

sachinswagh:
Require 'rpush' to resolve uninitialized constant error.


aried3r:
How can I reproduce this error?

sachinswagh:
Just install rpush and try to access on rails console.

Issue #439, opened 2018-06-27
Rpush Memory Leak


armahmoudi:
I'm using the CLI to run an Rpush daemon and i'm using Redis. It appears that the Rpush daemon has a memory leak which causes the process to consume more and more memory until the instance runs out of memory, then it causes notifications to be dropped (they stay in the queue but won't get a success/failure update), and sometimes get stuck in the internal queue.

These issues don't happen if there's memory available, but as soon as the host starts swapping the daemon has a hard time delivering notifications and eventually gets stuck.

aried3r:
Does Rpush use more and more memory or Redis? Redis is a known "issue", it was recently added as a gotcha in the README https://github.com/rpush/rpush#cleanup

armahmoudi:
@aried3r I'm talking about the application memory on the server, not Redis. I was aware of notifications not being cleaned up from Redis and had put a mechanism to destroy each notification on any success/failure reflection. That part works great as long as the application doesn't run out of memory.

armahmoudi:
@aried3r I'm trying to use Rpush for all of our Push Notifications at Tesla, and this is the main blocker to take it to production right now. Would appreciate any help I can get. Would be happy to contribute to the repository as well once we do find the issue.
Right now I'm trying to detect what's the cause of this growing memory usage on the process, and my suspicion is on the feeder and dispatchers.

aried3r:
Hey! Sorry for the silence, I was gone for the week.

Hmm, that's really strange. I haven't seen that behaviour before in production and you're the first to report it, but maybe you use it at a different scale at Tesla :)
But I am also not using Redis but Postgres for persistence with FCM. Could you describe your setup in more detail (Ruby, Rpush version etc)?

Sadly I don't have a lot of insight in how the feeders and dispatchers work exactly, all credit goes to @ileitch and contributors. But I'll monitor our own daemons more closely in the coming weeks to see if I experience the same issue. Sorry that I can't help more right now :\

armahmoudi:
@aried3r could it be that the [InterruptibleSleep](https://github.com/rpush/rpush/blob/master/lib/rpush/daemon/interruptible_sleep.rb#L5) is leaking dead threads?

aried3r:
It could be, but if I read the code and your heap dumps somewhat correctly, there's only `Kernel.sleep` "in" the threads, but it leaks the most memory. But judging from the ref counts, it might be modis (but might just be because of the number of notifications).

[...]

PR #441, opened 2018-07-09
Fix interruptible_sleep memory leak (#439)

armahmoudi:
I ran a memory benchmark using [michaelherold/benchmark-memory](https://github.com/michaelherold/benchmark-memory) and the it seems like ruby does retain the dead threads if the instance var is not explicitly set to `nil`

```ruby
require 'benchmark/memory'

class InterruptibleSleep
  def sleep(duration)
    @thread = Thread.new { Kernel.sleep(duration) }
    Thread.pass

    begin
      @thread.join
    rescue StandardError
      @thread = nil
    end
  end

  def stop
    @thread.kill if @thread
  rescue StandardError # rubocop:disable Lint/HandleExceptions
  ensure
    @thread = nil
  end
end

class InterruptibleSleep2 < InterruptibleSleep
  def sleep(duration)
    @thread = Thread.new { Kernel.sleep(duration) }
    Thread.pass

    begin
      @thread.join
    rescue StandardError
    ensure
      @thread = nil
    end
  end
end

class InterruptibleSleep3 < InterruptibleSleep
  def sleep(duration)
    @thread = Thread.new { Kernel.sleep(duration); nil }
    Thread.pass

    begin
      @thread.join
    rescue StandardError
      @thread = nil
    end
  end
end

SLEEP = 0.001 # seconds

is = InterruptibleSleep.new
is2 = InterruptibleSleep2.new
is3 = InterruptibleSleep3.new

Benchmark.memory do |x|
  x.report('IS Originial') { is.sleep(SLEEP) }
  x.report('IS 2') { is2.sleep(SLEEP) }
  x.report('IS 3') { is3.sleep(SLEEP) }

  x.compare!
end
```

Pay attention to `IS 2` which is the code that is on this PR, and it doesn't retain anything in memory.
```
Calculating -------------------------------------
        IS Originial     1.050M memsize (     1.050M retained)
                         4.000  objects (     4.000  retained)
                         0.000  strings (     0.000  retained)
                IS 2     1.050M memsize (     0.000  retained)
                         4.000  objects (     0.000  retained)
                         0.000  strings (     0.000  retained)
                IS 3     1.050M memsize (     1.050M retained)
                         4.000  objects (     4.000  retained)
                         0.000  strings (     0.000  retained)

Comparison:
        IS Originial:    1049712 allocated
                IS 2:    1049712 allocated - same
                IS 3:    1049712 allocated - same
```

### DIFF
--- a/lib/rpush/daemon/interruptible_sleep.rb
+++ b/lib/rpush/daemon/interruptible_sleep.rb
@@ -7,7 +7,8 @@ module Rpush
 
         begin
           @thread.join
-        rescue StandardError
+        rescue StandardError # rubocop:disable Lint/HandleExceptions
+        ensure
           @thread = nil
         end
       end
--- a/lib/rpush/version.rb
+++ b/lib/rpush/version.rb
@@ -2,7 +2,7 @@ module Rpush
   module VERSION
     MAJOR = 3
     MINOR = 2
-    TINY = 0
+    TINY = 1
     PRE = nil
 
     STRING = [MAJOR, MINOR, TINY, PRE].compact.join(".").freeze


aried3r:
Awesome! Nice find! Thanks for the pull request, I'll release a new version today.

PR #386, opened 2017-11-03
Added support for P8 certificate apple push notifications

mariannegru:
Apple push notifications using authentication tokens

http://help.apple.com/xcode/mac/current/#/dev54d690a66

PR #444, opened 2018-07-11
Update jwt to 2.1.0

### LINKED ISSUE #435: Update jwt to 2.x
I missed that in #386.

jsantos:
Fixes #435

I've also removed the references to `rails 5.2.0.beta2`, since a final version has been out for a while.

On `lib/rpush/daemon/apnsp8/token.rb` I just did some styling adjustments suggested by Rubocop (haven't really changed anything, but was consulting the class to confirm if anything had to be changed).

@aried3r please take a look since you were involved on the APNS P8 integration. Wonder if there shouldn't be some tests for this...

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -38,10 +38,10 @@ appraise "rails-5.1" do
 end
 
 appraise "rails-5.2" do
-  gem "activesupport", ">= 5.2.0.beta2"
+  gem "activesupport", ">= 5.2.0"
 
   group :development do
-    gem "rails", ">= 5.2.0.beta2"
+    gem "rails", ">= 5.2.0"
   end
 end
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,10 +1,10 @@
 PATH
   remote: .
   specs:
-    rpush (3.2.0)
+    rpush (3.2.2)
       activesupport
       ansi
-      jwt (~> 1.5.6)
+      jwt (~> 2.1.0)
       multi_json (~> 1.0)
       net-http-persistent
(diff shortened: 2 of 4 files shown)


aried3r:
> please take a look since you were involved on the APNS P8 integration

LGTM. But all credit goes to the awesome people in https://github.com/rpush/rpush/pull/386, I merely merged it :)

> Wonder if there shouldn't be some tests for this...

I would love to have some, but I sadly don't have the time or insight into this topic to do so. :\

sposinff:
Why was the update to jwt necessary? `oauth2 1.4.0` does not support `jwt 2.1.0`

jsantos:
@sposinff For the exact same reason: Incompatibility with other gems (which are up to date). Keeping `jwt` outdated might imply security risks for a myriad of users.

As I see, `oauth2` 2.0.0 already supports recent version of `jwt`, according to its gemspec, although it's not yet released. Probably another PR like this to the current version might solve the issue. 

You still can use `rpush` with older versions of `jwt` by specifying the previous version on your Gemfile:

`gem 'rpush', '3.2.2'`

Hope this helps! :)

sposinff:
I appreciate the desire to stay up-to-date with gems, especially for security reasons! Couldn't everyone be happy with this:

```s.add_runtime_dependency 'jwt', '>= 1.5.6'```

jsantos:
@schmid You're right! Since there were no required changes to the `jwt` usage from 1.5.6 to 2.1.0, I guess we can relax the dependency. 

I can issue another PR for this during the weekend.

PR #445, opened 2018-07-13
 Apns::Notification should be able to store multiple registration_ids

faucct:
Hello. Putting multiple `device_token`s in a single notification could ease things when sending large push campaigns.

### DIFF
--- a/lib/rpush/daemon/apns/delivery.rb
+++ b/lib/rpush/daemon/apns/delivery.rb
@@ -26,8 +26,10 @@ module Rpush
 
         def batch_to_binary
           payload = ""
-          @batch.each_notification do |notification|
-            payload << notification.to_binary
+          @batch
+            .each_notification.lazy.flat_map { |notification| notification.device_tokens.zip([notification].cycle) }
+            .each_with_index do |(device_token, notification), frame_id|
+            payload << notification.to_binary(frame_id: frame_id, device_token: device_token)
           end
           payload
         end
--- a/lib/rpush/client/active_model/apns/notification.rb
+++ b/lib/rpush/client/active_model/apns/notification.rb
@@ -74,18 +74,19 @@ module Rpush
             json
           end
 
-          def to_binary(options = {})
-            [*device_token, *registration_ids].each_with_index.map do |device_token, index|
-              frame_payload = payload
-              frame_id = options[:for_validation] ? 0 : send(options.fetch(:id_attribute, :id))
-              frame = ""
-              frame << [1, 32, device_token].pack("cnH*")
-              frame << [2, frame_payload.bytesize, frame_payload].pack("cna*")
-              frame << [3, 4, frame_id ^ index].pack("cnN")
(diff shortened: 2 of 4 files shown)


lib/rpush/daemon/apns/delivery.rb
```diff
@@ -26,8 +26,10 @@ def perform
 
         def batch_to_binary
           payload = ""
-          @batch.each_notification do |notification|
-            payload << notification.to_binary
+          @batch
```
vizvamitra: This will break APNs error handling functionality in [Rpush::Daemon::Dispatcher::ApnsTcp][1].

According to [The Binary Interface and Notification Format][2] document, "If you send a notification that is malformed or otherwise unintelligible, APNs returns an error-response packet and closes the connection. Any notifications that you sent after the malformed notification using the same connection are discarded, and must be resent. [...] The packet has [...] the notification identifier of the malformed notification".

That "notification identifier" will be equal to `frame_id` in [Rpush::Client::ActiveModel::Apns::Notification#to_binary][3]. Currently this ID equals to a `Rpush::Apns::Notification#id`, which helps `ApnsTcp` to determine the failed notification and mark it accordingly. But you've used an index as a frame_id here, so `ApnsTcp` won't be able to fail notifications correctly.

This is how I see the possible fix:

1. Encode both Notification.id and device_token into a frame_id
2. In `ApnsTcp`:
  - Parse notification_id returned by APNs to get the notification id and an index of the failed device_token
  - Find corresponding notification
  - Take all device_tokens after the failed one and create a new `Rpush::Apns::Notification` from them
  - Mark the old notification as failed

And you should also ensure that all the notifications sent after the failed in the same batch are marked as retryable.

Or maybe since a single notification can now contain multiple device tokens it is possible to switch off batch_deliveries for `Rpush::Daemon::Apns`?

[1]: https://github.com/rpush/rpush/blob/master/lib/rpush/daemon/dispatcher/apns_tcp.rb#L119
[2]: https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPhttps://github.com/rpush/rpush/blob/master/lib/rpush/client/active_model/apns/notification.rb#L77G/BinaryProviderAPI.html#//apple_ref/doc/uid/TP40008194-CH13-SW4
[3]: https://github.com/rpush/rpush/blob/master/lib/rpush/client/active_model/apns/notification.rb#L77

faucct: I thought that `Apns::Feedback` was the only error handling `Apns` had.
`Rpush::Daemon::Dispatcher::ApnsTcp#deliver_buffer` can help map `frame_id` to `notification` as it is used in `#handle_error` anyway.

aried3r:
@vizvamitra, @faucct, I'm not using APNs myself, so I need a bit of help here. Is this ready or are the issues mentioned in the review valid concerns?

faucct:
No, it is not valid.

Issue #451, opened 2018-08-28
undefined method `changed?' for nil:NilClass


robertziel:
Hi, after updating to `rails 5.2.1` I have a problem with assigning value for example:
```
app = Rpush::Apns::App.new
app.name = app_name
```

Which returns error:
`undefined method 'changed?' for nil:NilClass`
It worked before, and after a day of research I did not make much progress. It appears all the time when object related to `rpush` is initialised and assigned params to it like `Rpush::Apns::Notification.new(params)` and so on.

I have figured it out that the problem occurs in `activemodel/lib/active_model/attribute_mutation_tracker.rb:49`
While in `rails 5.1.0` this method is not called at all.

aried3r:
I'm using rpush with Rails 5.2.1 in production for a few weeks now and I'm not facing that issue. I also tried with a new Rails project using `rails new rpush_issue_451` and adding Rpush. I can create a new app just fine.

Are you sure it's Rpush? Are you by chance using anything to track changes of models, like the audited or paper_trail gem?

robertziel:
You are right. The problem is not related with `rpush` but with `rpush-redis` and `modis`, I am not sure which one but problem is when this method is called.
`https://github.com/ileitch/modis/blob/master/lib/modis/attribute.rb#L67`

robertziel:
It is related to this issue [https://github.com/ileitch/modis/issues/13](https://github.com/ileitch/modis/issues/13), so I think we may close this one. Sorry for problems :)

Benjamin-Dobell:
Ran into this also when setting up a new environment. Specifically:

```ruby
Rpush::Apns::App.new
app.certificate = cert # This line explodes
```

Whilst this is Modis issue, I think the ball may be back in Rpush's court as Modis isn't maintained anymore, so probably ought to be replaced in Rpush:
https://github.com/ileitch/modis/issues/13#issuecomment-386132174

soffes:
Agree with @Benjamin-Dobell. I had to switch to a different gem ([Houston](https://github.com/nomad/houston)).

benlangfeld:
A solution is proposed in https://github.com/ileitch/modis/pull/19.

PR #453, opened 2018-09-19
Fix for undefined method changed? seen in Rails 5.2 with Redis

mohitnatoo:
Fix for https://github.com/rpush/rpush/issues/451


benlangfeld:
I suggest that https://github.com/ileitch/modis/pull/19 is a superior solution to this problem, and that this PR should be closed.

aried3r:
Agreed. Thanks, @mohitnatoo for your PR anyway! :)

PR #455, opened 2018-09-24
Change apn_key column type from string to text

sonxurxo:


aried3r:
Hey! Thanks for the PR!

Sadly we cannot change existing migrations because people who are already on 3.2.0 usually can't re-run them. You'd have to create a new migration for the next version of Rpush.

sonxurxo:
Hi @aried3r, I was actually thinking about that when submitting the PR. Just updated it, I guess it will be version 3.2.4.

zedtux:
What's the next step for this PR?

aried3r:
@zedtux for me to merge and release it 😅

@sonxurxo, could you check out the failing tests? I restarted some of the jobs that just timed out, but there seems to be a problem with Rails 5.2 for some reason.

sonxurxo:
@aried3r I'll do, having a quick look I see that the failing test is

> rspec ./spec/unit/client/active_record/pushy/notification_spec.rb:52 # Rpush::Client::ActiveRecord::Pushy::Notification validates time_to_live should be integer

which apparently has nothing to do with the changes. Even, the test expects the model to *not* be valid due to an invalid data type in another attribute (`time_to_live` should be an integer and the test sets it to Float), but fails because it finds the model to be OK.

I'll try to have a look in detail later.

sonxurxo:
Just reviewed and I can't figure out what is wrong.

Running the specs in my dev machine using Ruby 2.3.7 and Rails 5.2 (like [this failing config](https://travis-ci.org/rpush/rpush/jobs/432802861/config)) does not reproduce that error thrown in Travis.

I'm quite stuck right now. Any hint?

aried3r:
Seeing as the failing tests are also on the master branch now, it shouldn't keep us from merging this any longer. Thanks for the PR!

zedtux:
Thank you @sonxurxo !

sonxurxo:
Great! Thank you all for this great library.

PR #456, opened 2018-10-12
Retry all GCM 5xx errors

rofreg:
According to the [GCM docs](https://developers.google.com/cloud-messaging/http-server-ref#error-codes), for errors in the 500-599 range, the sender "must retry later". Currently, rpush only retries these notifications if the error is specifically 500 or 503.

In production, we have seen this somewhat commonly with 502 Bad Gateway errors. This PR adds specific retry handling for 502 errors, as well as general purpose retry handling for other 5xx GCM errors.

### DIFF
--- a/lib/rpush/daemon/gcm/delivery.rb
+++ b/lib/rpush/daemon/gcm/delivery.rb
@@ -7,7 +7,7 @@ module Rpush
 
         host = 'https://fcm.googleapis.com'
         FCM_URI = URI.parse("#{host}/fcm/send")
-        UNAVAILABLE_STATES = %w(Unavailable InternalServerError)
+        UNAVAILABLE_STATES = %w(Unavailable BadGateway InternalServerError)
         INVALID_REGISTRATION_ID_STATES = %w(InvalidRegistration MismatchSenderId NotRegistered InvalidPackageName)
 
         def initialize(app, http, notification, batch)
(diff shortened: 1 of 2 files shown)


aried3r:
Hey @rofreg, thanks, that looks good!

I'll try to have a look soon. But I'm wondering why Travis didn't pick up this PR 🤔

aried3r:
Ok, seems Travis changed their settings at some point. I activated building and testing PRs again, hopefully closing and opening will [trigger a build](https://github.com/travis-ci/travis-ci/issues/576#issuecomment-65317915).

rofreg:
Updated! I believe that the one remaining test failure is a failure on `master` branch, as the same test appears to be failing for all other PRs too (e.g. #455). I might try to look into that later if time permits – if so, I'll open a separate PR for it.

aried3r:
Thanks @rofreg!



> I might try to look into that later if time permits – if so, I'll open a separate PR for it.

That would be really awesome! It's curious, since it used to work and we use it in production at work with the latest Rails 5.x version 🤔

aried3r:
Thanks for digging into this and the PR! Merging, since the errors on CI aren't caused by this PR.

aried3r:
Followup, might be https://github.com/rails/rails/issues/33651, it's just strange that I can't reproduce it locally.

PR #460, opened 2018-10-24
Add thread-id for APNs support

gerard-morera:
Hello, this PR adds support for Apple Push Notification service `thread-id` field.

> thread-id: An app-specific identifier for grouping related notifications. This value corresponds to the threadIdentifier property in the UNNotificationContent object.

For more information check [Apple notifications docs](https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification
).

### DIFF
--- a/spec/unit/client/active_record/apns/notification_spec.rb
+++ b/spec/unit/client/active_record/apns/notification_spec.rb
@@ -309,3 +309,16 @@ describe Rpush::Client::ActiveRecord::Apns::Notification, "multi_json usage" do
     end
   end
 end if active_record?
+
+describe Rpush::Client::ActiveRecord::Apns::Notification, 'thread-id' do
+  let(:notification) { Rpush::Client::ActiveRecord::Apns::Notification.new }
+
+  it 'includes thread-id in the payload' do
+    notification.thread_id = 'THREAD-ID'
+    expect(notification.as_json['aps']['thread-id']).to eq 'THREAD-ID'
+  end
+
+  it 'does not include thread-id in the payload if not set' do
+    expect(notification.as_json['aps']).to_not have_key('thread-id')
+  end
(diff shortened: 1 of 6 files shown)


spec/unit/client/active_record/apns/notification_spec.rb
```diff
@@ -309,3 +309,16 @@
...
+describe Rpush::Client::ActiveRecord::Apns::Notification, 'thread-id' do
+  let(:notification) { Rpush::Client::ActiveRecord::Apns::Notification.new }
+
+  it 'includes thread-id in the payload' do
+    notification.thread_id = 'THREAD-ID'
+    expect(notification.as_json['aps']['thread-id']).to eq 'THREAD-ID'
+  end
+
+  it 'does not include thread-id in the payload if not set' do
+    expect(notification.as_json['aps'].key?('thread-id')).to be_falsey
```
aried3r: I think there's a `has_key?` matcher which we could use.

lib/generators/templates/rpush_3_2_5_updates.rb
```diff
@@ -0,0 +1,9 @@
+class Rpush325Updates < ActiveRecord::VERSION::MAJOR >= 5 ? ActiveRecord::Migration[5.0] : ActiveRecord::Migration
```
aried3r: Could you maybe change it to rpush_3_3_0_updates? Since it's kind of a feature, we might go for a 3.3.0 release instead of a patch release.

aried3r:
Hey @gerard-morera! Thanks for your contribution! I left a few remarks, but otherwise it looks good. Thanks!

gerard-morera:
@aried3r thanks for your review! I have applied the requested changes.

PR #463, opened 2018-11-04
remove validation of 64-characters length from device_token

### LINKED ISSUE #462: Validity check failed: Device token is invalid
Hi.. I faced a problem when trying to create a Rpush::Apns::Notification because the device_token from an user was longer than 64 characters and according to iOS documentation (https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/APNSOverview.html#//apple_ref/doc/uid/TP40008194-CH8-SW14):

> APNs device tokens are of variable length. Do not hard-code their size. 

So the thing is that in the generators templates you put a limit of 64 characters for this attribute. In the case that I saw in my project, an user has a device_token with 108 characters. He installed the app but he doesn't receive the notifications because the validation fails. I think it's better to remove the length restriction, according to the documentation link above.

I already created a local migration to change that in the Rpush::Notification table of my project but Rpush has this validator that doesn't allow me to save it :https://github.com/rpush/rpush/blob/master/lib/rpush/client/active_model/apns/device_token_format_validator.rb

chrisokamoto:
This PR migrates the device_token from string(64) to just string - 255 characters. My first thought was to migrate the field to a text field, but then I saw that we have indexes on device_token and a very big field could be a problem. And with text field I actually couldn't run the migration. :(

Fixes #462

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ app.save!
 ```ruby
 n = Rpush::Apns::Notification.new
 n.app = Rpush::Apns::App.find_by_name("ios_app")
-n.device_token = "..." # 64-character hex string
+n.device_token = "..." # hex string
 n.alert = "hi mom!"
 n.data = { foo: :bar }
 n.save!
@@ -97,7 +97,7 @@ app.save!
 ```ruby
 n = Rpush::Apns::Notification.new
 n.app = Rpush::Apnsp8::App.find_by_name("ios_app")
-n.device_token = "..." # 64-character hex string
+n.device_token = "..." # hex string
 n.alert = "hi mom!"
 n.data = { foo: :bar }
 n.save!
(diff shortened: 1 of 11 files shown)


aried3r:
Hey! Thanks for your PR, just some minor notes. I think we should add a new migration instead of changing existing ones so people who already ran all the migrations will also profit from this change, not just people who install rpush for the first time.

A similar PR that changed limits of certain columns can be found here in https://github.com/rpush/rpush/pull/455

chrisokamoto:
@aried3r all right!! :) I'll create the new migrations then. Thanks!
I also saw in this other PR that you cited that there are some failing tests like the ones that I got on this PR here in master.. I was wondering why they were failing in my PR since I didn't touch anything concerned to _time_to_live_ , so the problem is with master branch. :( So TravisCi is red here, but I wrote the tests to coverage my changes.

chrisokamoto:
@aried3r I wrote the migration, can you check if it's ok for you now? The failing tests are the ones that were already failing in master. :(

chrisokamoto:
@aried3r did you have a chance to look into this PR?

aried3r:
@chrisokamoto, looking at it now, thanks for your contribution! 

I rebased and squashed it and once Travis CI finishes and everything is fine I'll merge and hopefully push a release today :)

PR #470, opened 2019-01-16
APNS p8 improvements

rofreg:
#### 1. Stop logging all apnsp8 requests as warnings

The apnsp8 adapter currently logs the full request data for each individual notification with a log level of `warn`, which doesn't appear to be done for other service adapters in rpush. This leads to a lot of unnecessary logs at scale, and it could potentially expose private payload data in server logs in an undesired way.

This logging line was added in the initial PR that added apnsp8 support (#386, https://github.com/rpush/rpush/commit/3bc1929b98c0f29dd7fbca1ba0c8531fe879120e) – I suspect that it was added to help with initial implementation and debugging, and that it was left in as an oversight. I think it should be safe to remove, but please let me know if I've missed something. (And many thanks to the initial implementor for adding apnsp8 support!)

#### 2. Fix a memory leak

We send a large volume of APNS notifications per day, and when we switched from using `Rpush::Apns::Notification` to `Rpush::Apnsp8::Notification`, we saw a dramatic increase in memory usage by the daemon. Memory usage scales linearly as you send more messages – sending 100K messages made our memory usage grow by roughly 500MB. Eventually our rpush instance was forced to use swap memory to keep running, and deliveries ground to a halt.

We're not sure if this PR fully fixes the problem, but it was the first obvious leak that we could find. A new proc is allocated for every individual notification, and never released. We intend to run this modified version of rpush in production, and to continue investigating if we see any further issues.

### DIFF
--- a/lib/rpush/daemon/apnsp8/delivery.rb
+++ b/lib/rpush/daemon/apnsp8/delivery.rb
@@ -14,8 +14,6 @@ module Rpush
           @batch = batch
           @first_push = true
           @token_provider = token_provider
-
-          @client.on(:error) { |err| mark_batch_retryable(Time.now + 10.seconds, err) }
         end
 
         def perform
--- a/lib/rpush/daemon/apnsp8/delivery.rb
+++ b/lib/rpush/daemon/apnsp8/delivery.rb
@@ -14,11 +14,11 @@ module Rpush
           @batch = batch
           @first_push = true
           @token_provider = token_provider
+
+          @client.on(:error) { |err| mark_batch_retryable(Time.now + 10.seconds, err) }
         end
 
         def perform
-          @client.on(:error) { |err| mark_batch_retryable(Time.now + 10.seconds, err) }
-
           @batch.each_notification do |notification|
             prepare_async_post(notification)
           end
--- a/lib/rpush/daemon/apnsp8/delivery.rb
+++ b/lib/rpush/daemon/apnsp8/delivery.rb
@@ -43,7 +43,6 @@ module Rpush
           response = {}
 
           request = build_request(notification)
-          log_warn(request)
           http_request = @client.prepare_request(:post, request[:path],
             body:    request[:body],
             headers: request[:headers]
--- a/lib/rpush/daemon/dispatcher/apnsp8_http2.rb
+++ b/lib/rpush/daemon/dispatcher/apnsp8_http2.rb
@@ -16,11 +16,11 @@ module Rpush
 
           url = URLS[app.environment.to_sym]
           @client = NetHttp2::Client.new(url, connect_timeout: DEFAULT_TIMEOUT)
+          @client.on(:error) {|error| reflect(:error, error)}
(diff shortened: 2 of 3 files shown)


aried3r:
Hey Ryan, Awesome PR! I'll try to have a look but since I don't use APNS nor have written the code for it I probably can't really say much about it. I'll try to have this merged soon.

Maybe @mariannegru can take a look as well if there's time. No obligations of course. :)

lib/rpush/daemon/apnsp8/delivery.rb
```diff
@@ -14,11 +14,11 @@ def initialize(app, http2_client, token_provider, batch)
           @batch = batch
           @first_push = true
           @token_provider = token_provider
-        end
 
-        def perform
           @client.on(:error) { |err| mark_batch_retryable(Time.now + 10.seconds, err) }
```
aried3r: I saw that this is also the case for regular `apns2`, as in, it's called in `perform`. Should we also change it there?

My question would also be, have you verified this works as intended if it's called in `initialize`? This construct (`@client.on`) is only used in `apns2` and `apnsp8` and nowhere else in rpush which makes me think it was copied over from `apns` during the `apnsp8` implementation, but I'm not sure exactly how it came to be.


https://github.com/rpush/rpush/blob/a8e7798de16c645b0185fb81950fdac49dac68d9/lib/rpush/daemon/apns2/delivery.rb#L18

aried3r: Hmm, I think having a partial fix is also welcome for others. I'm just not sure about the `client.on` stuff.

Have you seen #439 and https://github.com/ileitch/modis/issues/14 in case you are also using modis?

jhottenstein: Thanks for the feedback @aried3r 

I've dug into this issue a little more (I work with @rofreg) and I think I've identified the location of this memory leak (and also the memory leak associated with #439 (i don't believe modis is actually involved).  I'll write up my findings in #439 and push a solution to this PR.  

We can add the change to apns2 to this PR or open a new one if you prefer.

aried3r: Awesome, and thanks for your write-down! I'll try to have a look.



> We can add the change to apns2 to this PR or open a new one if you prefer.

Whatever is easier for you. If it's in individual commits, why not in this one.

jhottenstein: @aried3r, it was easier for me to split this PR into a couple PRs so I have opened #474, #475 and #476 

#474 and #475 are currently in our production instance; we don't use apns2 so I can't test #476.

I can collapse them back to this PR if you prefer but otherwise they are ready to be merged.

aried3r: Awesome! Makes it much easier to review and follow. I guess the new PRs supersede this PR?

aried3r:
Closing. Superseded by #474, #475, #476.

Issue #413, opened 2018-02-15
Rpush with systemctl


spaghetticode:
I think the readme could be clearer about how to make rpush work nice with `systemctl`: unless you add the -f flag the command `rpush start` will not work correctly when supervised by systemctl. 

This happens because, as you correctly wrote, rpush deamonizes by default when started with `rpush start`. 

Having an explicit suggestion about `rpush start -f` for not daemonizing the service (and playing nice with systemctl) seems to me a good improvement to the readme.

I can open a PR for this, if it helps.

Of course, thank you very much for this great gem!

aried3r:
@spaghetticode, I don't use rpush using systemctl but was actually planning to look into that. I – and I think many others as well – would be really grateful if you could improve the documentation on that.

elieteyssedou:
I think we can close this issue because this argument exists : https://github.com/rpush/rpush/blob/master/lib/rpush/cli.rb#L19

Enjoy !

(may be it would be nice to update the documentation)

aried3r:
Done with #478, thanks @elieteyssedou!

elieteyssedou:
You are welcome ! 😁

PR #478, opened 2019-02-08
Update readme, add foreground (-f) argument

elieteyssedou:
#413

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,13 @@ $ cd /path/to/project
 $ rpush start
 ```
 
+#### As a foreground process
+
+```sh
+$ cd /path/to/project
+$ rpush start -f
+```
+
 #### On the command-line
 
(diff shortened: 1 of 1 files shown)


aried3r:
Awesome, thank you!

Btw, if you happen to have a working systemd configuration, I'd be happy if you could share it.

There's https://github.com/rpush/rpush/blob/master/examples/rpush_systemd.service, but I don't know if that's the best way to do it.

elieteyssedou:
@aried3r I am using Upstart, not Systemd, sorry ! 😞

aried3r:
Oh, we also have an example for that, but it's 4 years old. Maybe you could take a look at it or replace it with yours? No obligations of course!

https://github.com/rpush/rpush/blob/master/examples/upstart-rpush.conf

PR #486, opened 2019-03-18
Fail gracefully when a Modis notification no longer exists

rofreg:
This makes behavior more consistent with the ActiveRecord adapter. If a service attempts to call #mark_ids_retryable or #mark_ids_failed and includes an ID that no longer exists in the datastore, the adapter will simply skip that record and continue, rather than crashing. (Currently, the adapter crashes, which can cause delivery issues with other unrelated notifications.)

### DIFF
--- a/lib/rpush/daemon/store/redis.rb
+++ b/lib/rpush/daemon/store/redis.rb
@@ -17,7 +17,7 @@ module Rpush
           limit -= retryable_ids.size
           pending_ids = limit > 0 ? pending_notification_ids(limit) : []
           ids = retryable_ids + pending_ids
-          ids.map { |id| Rpush::Client::Redis::Notification.find(id) }
+          ids.map { |id| find_notification_by_id(id) }.compact
         end
 
         def mark_delivered(notification, time, opts = {})
@@ -49,7 +49,12 @@ module Rpush
         end
 
         def mark_ids_failed(ids, code, description, time)
-          ids.each { |id| mark_failed(Rpush::Client::Redis::Notification.find(id), code, description, time) }
+          ids.each do |id|
+            notification = find_notification_by_id(id)
+            next unless notification
+
+            mark_failed(notification, code, description, time)
+          end
         end
 
         def mark_retryable(notification, deliver_after, opts = {})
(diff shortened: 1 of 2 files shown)


aried3r:
Hey, thanks for the contribution!

Is there maybe a way we can add a small test for this to document the behavior?

rofreg:
Tests added!

aried3r:
Awesome, thanks! I wonder if this constitutes as a breaking change 🤔. People might rely on the exception being thrown to do their own error handling. WDYT wrt semver?

rofreg:
Fair question! Personally, I think this should be considered a bugfix rather than a breaking change.

The `mark_ids_XXX` methods appear to only be used inside the [ApnsTcp dispatcher](https://github.com/rpush/rpush/blob/ed20cf4425a757859f9dde35dd4096f15beb1317/lib/rpush/daemon/dispatcher/apns_tcp.rb#L124). They throw an exception that prevents [the rest of the method](https://github.com/rpush/rpush/blob/ed20cf4425a757859f9dde35dd4096f15beb1317/lib/rpush/daemon/dispatcher/apns_tcp.rb#L125-L132) from running (which prevents reflection, retries, etc), but that exception is then [rescued a few levels up by the dispatcher itself](https://github.com/rpush/rpush/blob/ed20cf4425a757859f9dde35dd4096f15beb1317/lib/rpush/daemon/dispatcher/apns_tcp.rb#L86-L88), so the exception doesn't seem to bubble up enough for a person to do their own error handling.

aried3r:
Great points, thanks for digging that up! I'll try to have a look at it soon and release a version with a bugfix.

aried3r:
Hey! Sorry to only reach out to you now, I've been gone for a week.

Something's strange with the tests, in general, not just yours.

584 tests on master: https://travis-ci.org/rpush/rpush/jobs/506269453#L2460
584 tests on this branch: https://travis-ci.org/rpush/rpush/jobs/508562778#L2449

Even though you added two tests. Maybe I'm doing something wrong here, but I could also not make them run using:

```console
$ be appraisal rails-5.2 rspec spec/unit/daemon/store/redis_spec.rb
Finished in 0.00218 seconds (files took 2.57 seconds to load)
0 examples, 0 failures
```

Adding something like the following

```Ruby
# redis_spec.rb
it 'does nothing' do
  fail 'this should fail'
end
```

does not result in an additional test and no failed test when running:

```console
$ be appraisal rails-5.2 rspec spec/unit/daemon/store/
Finished in 0.89417 seconds (files took 2.57 seconds to load)
75 examples, 0 failures
```

aried3r:
🤦‍♂️ maybe I've been gone too long, haha. It's a redis test and I'm not using `CLIENT=redis`. Nevermind, I'll check again.

rofreg:
Haha, no worries! I made the same mistake 🙂 Thanks for your work!

PR #502, opened 2019-05-16
Add support for critical alerts

servn:
According to https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/generating_a_remote_notification#2990112
Apns::Notification sound attribute could be a Dictionary for critical alerts


drn:
Hi @servn - apologies for months-late reply to this PR.

This looks great! I'll merge this in and update the changelog.

Are you interested in helping maintain rpush?

aried3r:
I believe this adds too many migrations (only one is needed) and doesn't add the 4.1.2 migration to the test suite.

I think these issues should be addressed before releasing.

drn:
@aried3r Should have waited for your review. I can take on addressing in the next few days if @servn is unable to

PR #506, opened 2019-06-26
Add mutable_content support for GCM

hon3st:
https://firebase.google.com/docs/cloud-messaging/http-server-ref

Currently it doesn't send `mutable_content` option to firebase server


aried3r:
Thank you!

PR #513, opened 2019-07-31
 Handle GCM reflections in batches.

balbesina:
Since we push messages to GCM in a batch and the response comes with each registration_id delivery status, there is no need to reflect each device separately. This feature allows to receive all successful and failed statues per single reflection.
```ruby
# example usage
Rpush.configure do |config|
  config.gcm_batch_reflections = true
end
Rpush.reflect do |on|
  on.gcm_delivered_to_recipients do |notification, successes_map|
    success_values = successes_map.keys.map { |token| [notification.id, token] }
    DeviceDeliveryStatus.import %i[notification_id device_token], success_values
  end
end
```

### DIFF
--- a/lib/rpush/daemon/gcm/delivery.rb
+++ b/lib/rpush/daemon/gcm/delivery.rb
@@ -72,10 +72,17 @@ module Rpush
         end
 
         def handle_successes(successes)
-          successes.each do |result|
-            reflect(:gcm_delivered_to_recipient, @notification, result[:registration_id])
-            next unless result.key?(:canonical_id)
-            reflect(:gcm_canonical_id, result[:registration_id], result[:canonical_id])
+          return if successes.blank?
+
+          if Rpush.config.gcm_batch_reflections
+            successes_map = successes.map { |result| [result[:registration_id], result[:canonical_id]] }.to_h
+            reflect(:gcm_delivered_to_recipients, @notification, successes_map)
+          else
+            successes.each do |result|
+              reflect(:gcm_delivered_to_recipient, @notification, result[:registration_id])
+              next unless result.key?(:canonical_id)
+              reflect(:gcm_canonical_id, result[:registration_id], result[:canonical_id])
+            end
           end
         end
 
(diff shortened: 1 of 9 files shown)


aried3r:
Hey! Thanks for your PR, I think it makes a lot of sense :)

Could you please add tests for it? There's quite a lot going on, new conditionals and changes in behavior. I think it'd be best some tests would document this.

balbesina:
I'll find some time on the upcoming weekend for that.

balbesina:
Added same option for notifications: :notification_batch_reflections
Added spec coverage.

balbesina:
Please provide acceptable resolution for rubocop's "block too long" of delivery_spec:
1. add disable comment for block length in file
2. extend block length for spec folder in .rubocop.yml
3. extract new examples to separate delivery2_spec

aried3r:
Hey! Sorry I've neglected this PR for a while. @rpush/maintainers, kindly asking for a second pair of eyes for review.

lib/rpush/daemon/gcm/delivery.rb
```diff
@@ -72,10 +72,17 @@ def process_response(response)
...
 
         def handle_successes(successes)
-          successes.each do |result|
-            reflect(:gcm_delivered_to_recipient, @notification, result[:registration_id])
-            next unless result.key?(:canonical_id)
-            reflect(:gcm_canonical_id, result[:registration_id], result[:canonical_id])
+          return if successes.blank?
+
+          if Rpush.config.gcm_batch_reflections
+            successes_map = successes.map { |result| [result[:registration_id], result[:canonical_id]] }.to_h
```
AlexTatarnikov, line 78: It looks like potential memory issue here and for `failures_map` in `handle_errors` as well. 

`successes` can hold thousands of objects and `map` will create a new array in memory, the same `to_h` does. 

You can use something like `inject` or `each_with_object ` in order to reduce the number of iterations and objects but still not sure if it is good idea "double" the memory consumption.

benlangfeld:
GCM has been shut down and this is stale.

PR #515, opened 2019-08-13
Stale bot config change

aried3r:
Hey @drn, thanks for moving this along!

I have two suggested changes, what do you think?

I'm also not sure if we should exempt pull requests.


drn:
Makes sense & looks good to me. I think that PRs would be fine to include - I suspect after a year the changes would get out of date. No strong opinions here, so up to you!

Going to merge this in now so that the bot does not accidentally close any `pinned` issues / PRs.

PR #519, opened 2019-09-03
Explicitly use Rails 6.0.0

jsantos:
Rails 6.0.0 was released few weeks ago: https://weblog.rubyonrails.org/2019/8/15/Rails-6-0-final-release/

I've also fixed some deprecation warnings I saw while running the test
suite: Replacing `.update_attributes` with `.update`.


aried3r:
Hey, thanks for your contribution!

I just fixed our Travis setup, so if you could rebase your branch we can check if everything is in order with your PR, but I assume there'll be no problems. :)

jsantos:
Just rebased it!

aried3r:
Hey! Seems there's some issues with the `update` and `update_attributes` with Redis. Could you maybe look into it what's causing it? Thanks!

aried3r:
Maybe it's Modis that doesn't have `.update`? https://github.com/ileitch/modis/blob/master/lib/modis/persistence.rb#L156-L169

jsantos:
@aried3r Yeah just realized this wasn't an actual `ApplicationRecord`, so this might compromise the Redis implementation. Guess we can live with the warnings for now, will create an issue for Modis about this.

Will remove the changes related to `update_attributes`.

Issue #529, opened 2019-10-01
Update RSpec


aried3r:
I tried this a few times, but even updating from 3.4.0 to 3.5.0 fails for me and I have no idea why. Maybe I was just unlucky and now it will just work. :)

https://github.com/rpush/rpush/pull/429
https://github.com/rpush/rpush/pull/482

yihyang:
@aried3r I have created a PR for this: https://github.com/rpush/rpush/pull/530

PR #530, opened 2019-10-01
Updated RSpec version

yihyang:
Fixed https://github.com/rpush/rpush/issues/529

Changes
- Updated rspec version


aried3r:
Thanks for the PR! As you can see, sadly all Redis test runs don't work if RSpec is upgraded, sadly I don't know why exactly.

yihyang:
I see, I think there are already some errors on the current versions. However upgrading the RSpec caused it to fail. GIve me sometime to have a look if you are ok.

Check the existing PR: https://travis-ci.org/rpush/rpush/jobs/590578647
There are some error messages that actually occurred but was not enough to cause it to fail

aried3r:
That would be awesome, thank you!

benlangfeld (changes requested):
Hi @yihyang , thanks for the contribution!

This includes a downgrade of Modis, which I don't think is desirable, and has some merge conflicts. Could you maybe resolve those two?

aried3r:
@benlangfeld, just for a little background, I tried a few times updating RSpec, but always failed due to obscure failures that IMO should not have been caused by a switch from 3.4 to 3.5. At least I couldn't find anything in the changelog of rspec that would suggest so.

I didn't want to invest the time to maybe even git-bisect through v3.4.0 to v3.5.0, but with rspec 4 being a thing, I assume this might make maintenance easier along the way.

aried3r:
I'm closing this but will reopen if there's new activity.

PR #541, opened 2019-12-11
Fix Rails 6.1 related deprecation warnings

dsantosmerino:
**Blocked by https://github.com/rpush/modis/pull/27**

We are upgrading to Rails 6 our main application and we have noticed some deprecation warnings coming from Rpush code. With this PR those deprecation warnings should be fixed.

- `#update_attributes` -  It was just an alias of `#update`, now deprecated.
- `case_sensitive: true` - Uniqueness validator will no longer enforce case sensitive comparisons.


aried3r:
Thank you for the contribution!

PR #547, opened 2020-01-02
Drop support for Rails 4.2

aried3r:
Rails 4.2 has reached EOL with the release of Rails 6. It has also
caused additional work recently to make CI work properly. With this, we
can remove the workarounds and focus on future version.


rpush.gemspec
```diff
@@ -35,7 +35,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency 'net-http-persistent'
   s.add_runtime_dependency 'net-http2', '~> 0.14'
   s.add_runtime_dependency 'jwt', '>= 1.5.6'
-  s.add_runtime_dependency 'activesupport'
+  s.add_runtime_dependency 'activesupport', '>= 5.0'
```
aried3r: We can discuss this change. Without this, people can still install future rpush versions on Rails 4.2, but there's no guarantee it will and I'd rather prevent issues from being opened for unsupported Rails versions. WDYT?

drn: 👍 Makes sense to me. Worth noting that in changelog + doing a minor version update

aried3r: Thanks for the feedback!
I'll leave this PR open another week or so since some people might be on holidays still. :)

PR #528, opened 2019-09-27
Improve test coverage

jhottenstein:
- bump .ruby_version to 2.6.4
- fix simplecov deprecation warning
- use shared examples to share tests between spec/unit/client/active_record and spec/unit/client/redis 
- fix two bugs in redis client classes
- use skipped tests to highlight two potential bugs in redis client classes

PR #561, opened 2020-09-23
Allow Apns2 payloads to be up to 4096 bytes

### LINKED ISSUE #545: Some notifications incorrectly marked as invalid due to payload size
**Describe the bug**

* Some notifications incorrectly marked as invalid due to payload size.
* APNS v1 imposes a validation payload size limit of 2kb. On APNS2, the payload limit increases to 4kb.
* When using APNS2, Rpush is validating the notification payload size based on APNS v1 size limits (2kb), rather than APNS v2 size limits (4kb).

**To Reproduce**

1. Create an APNS2 notification, where the `record.payload.bytesize` is greater than 2kb, but less than 4kb.
1. The notification is not sent, because it has the [following validation error](https://github.com/rpush/rpush/blob/5405439a83b87cd9e28999795bdfb8bf71c90c51/lib/rpush/client/active_model/apns/binary_notification_validator.rb#L6-L10): "APN notification cannot be larger than 2048 bytes. Try condensing your alert and device attributes."

**Expected behavior**

The notification payload should be valid, because APNS2 imposes a notification payload size of 4kb.

**Actual Behavior**

The notification payload is marked as invalid, because the `record.payload.bytesize` is greater than 2kb (even though APNS2 has a size limit of 4kb).

**Logs and other output**

N/A

**System configuration (please complete the following information):**

 - OS: macOS 
 - OS version: 10.14.6
 - Ruby version: ruby-2.5.0
 - Rails version: Rails 5.1.6.2
 - Rpush version: Rpush 3.0.2

**Additional context**

* Here's a [Link to Apple's docs](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html) describing the payload sizes for APNS v1 and v2.

This section of the docs *might* be relevant to the issue I'm encountering:

> If you are using the legacy APNs binary interface to send notifications instead of an HTTP/2 request, the maximum payload size is 2KB (2048 bytes)

* Is it possible that I'm using the legacy APNs binary, and that's why the 2kb limit is being imposed? In that case, this may not be an Rpush bug at all, and I'd be curious how to change my Rpush config to stop using the legacy binary interface.

**Workarounds (and potential solutions)**

* I've patched my `config/initializers/rpush.rb`, to override [the `MAX_BYTES` constant](https://github.com/rpush/rpush/blob/5405439a83b87cd9e28999795bdfb8bf71c90c51/lib/rpush/client/active_model/apns/binary_notification_validator.rb#L6) so it's 4kb:
`Rpush::Client::ActiveModel::Apns::BinaryNotificationValidator::MAX_BYTES = 4.kilobytes` (obviously, I'd prefer not overriding a constant, especially one set in a gem).
* Locally, I've tested replacing [this line](https://github.com/rpush/rpush/blob/master/lib/rpush/client/active_model/apns/notification.rb#L18) with  `validates_with Rpush::Client::ActiveModel::PayloadDataSizeValidator, limit: 4096`. This change *does* seem to resolve the issue. If a PR along these lines would be appropriate, please let me know.

benlangfeld:
https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html

Fixes #545


benlangfeld:
There's a bunch of places here where Redis and ActiveRecord behaviour diverges, particularly around serialization. I patched over what I could, and skipped tests where I couldn't, but this should be looked at again in a separate change. Here I just wanted to correct the total lack of test coverage when running on Redis in order to be sure I wasn't introducing regressions.

lib/rpush/client/active_model/apns2/notification.rb
```diff
@@ -3,6 +3,20 @@ module Client
     module ActiveModel
       module Apns2
         include Rpush::Client::ActiveModel::Apns
+
+        module Notification
+          MAX_PAYLOAD_BYTESIZE = 4096
```
gregblake, line 8: This change is key to squashing the behavior described in https://github.com/rpush/rpush/issues/545, and I'd like to provide some additional context to make reviewing a bit easier for other reviewers:

Prior to this PR, the APNS2 notification payload was incorrectly marked as invalid on notifications where the payload byte size was _greater than_ 2kb, but _less than_ 4kb. In other words, Apple has a [size limit of 4kb for APNS2](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html#:~:text=The%20maximum%20size%20of%20the,size%20is%205KB%20(5120%20bytes)), but rpush would mark any notifications with a payload greater than 2kb as invalid.

lib/rpush/client/active_model/apns/notification.rb
```diff
@@ -7,15 +7,23 @@ module Notification
           APNS_PRIORITY_IMMEDIATE = 10
           APNS_PRIORITY_CONSERVE_POWER = 5
           APNS_PRIORITIES = [APNS_PRIORITY_IMMEDIATE, APNS_PRIORITY_CONSERVE_POWER]
+          MAX_PAYLOAD_BYTESIZE = 2048
```
gregblake, line 10: This is the correct maximum payload size that Apple enforces on APNS 1:

For other reviewers, here's the [relevant documentation](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html#:~:text=The%20maximum%20size%20of%20the,size%20is%205KB%20(5120%20bytes)):

> If you are using the legacy APNs binary interface to send notifications instead of an HTTP/2 request, the maximum payload size is 2KB (2048 bytes)

gregblake (approved):

benlangfeld:
I'm looking at potentially merging https://github.com/rpush/rpush/pull/528 before this one.

benlangfeld:
@aried3r I removed the backward incompatibility you were concerned about here, and reduced the diff significantly by merging https://github.com/rpush/rpush/pull/528.

aried3r:
Looking good, thank you! I see your plan for this one is v5.0.1 and some other PRs are for 6.0.

For 6.0 I'm okay to harmonize the default values. As long as we document what we're changing, people should be okay migrating I assume. It's hard to tell if more people rely on there being default values, like for the sound, or if people are overriding them. But as long as we leave a migration note, either option would be fine.

spec/unit/client/redis/apns2/notification_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  it_behaves_like 'Rpush::Client::Apns::Notification'
+
+  it "should validate the length of the binary conversion of the notification" do
+    notification = described_class.new
+    notification.device_token = "a" * 108
+
+    notification.alert = "a" * 4095
+    expect(notification.errors[:base]).to be_empty
+
+    notification.alert = "a" * 4096
```
aried3r, line 15: Unless I'm missing something, isn't 4096 still allowed and 4097 would be a problem? From the documentation:

> For regular remote notifications, the maximum size is 4KB (4096 bytes)

For me that reads as ≤4096.

benlangfeld: You're right, sorry, fix on the way.

benlangfeld: Pointing out the error was much appreciated, it highlighted that the test was utterly deficient. It should now be correct.

benlangfeld:
@aried3r If it's possible to cut 5.0.1 or 5.1.0 once CI finally provides some feedback on this, I'd be very grateful.

aried3r:
Version 5.1.0 has just been pushed to rubygems.

PR #563, opened 2020-09-24
Improve DB reconnection for big tables

AlexTatarnikov:
Using `.count` for checking if DB connected can produce significant load on big tables. This commit simply replaces it by `.exists?`.

Screenshot 2020-09-24 at 12 36 04

image (1)


aried3r:
Good find, awesome! Thank you!

PR #564, opened 2020-09-24
Resume APNS2 delivery when async requests timeout

### LINKED ISSUE #477: APNSp8 dispatcher occasionally hangs
We are using the apnsp8 adapter to send a large amount of push notifications and we will occasionally (~6 times/day) get a batch that will error out after 7200 seconds (when the remote connection is closed).

I believe this is the same problem reported in #354

I think setting a timeout around client.join (or inside client.join is the appropriate solution).

benlangfeld:
Sometimes an async APNSv2 request's response will not be received.

Previously, rpush would wait indefinitely for this and continue pulling notifications out of the pending queue in redis to queue them in memory. Since the response never comes, this in-memory queue is not worked any further, the process gets stuck and will not cleanly shut down, requiring a SIGKILL. Those notifications queued in memory will forever remain in redis in a checked-out state such that the new rpush instance that replaces the hung one won't pick them up, and they'll thus never be sent.

This change not only retries sending the batch which was never fully confirmed as having been sent, but also ensures that processing of the in-memory queue continues past that batch, which in turn allows a shutdown to proceed smoothly, if slowly.

Fixes #477.

Based on https://github.com/ostinelli/net-http2/pull/43

### DIFF
--- a/lib/rpush/daemon/apns2/delivery.rb
+++ b/lib/rpush/daemon/apns2/delivery.rb
@@ -7,6 +7,7 @@ module Rpush
 
       class Delivery < Rpush::Daemon::Delivery
         RETRYABLE_CODES = [ 429, 500, 503 ]
+        CLIENT_JOIN_TIMEOUT = 60
 
         def initialize(app, http2_client, batch)
           @app = app
@@ -20,7 +21,11 @@ module Rpush
           end
 
           # Send all preprocessed requests at once
-          @client.join
+          @client.join(timeout: CLIENT_JOIN_TIMEOUT)
+        rescue NetHttp2::AsyncRequestTimeout => error
+          mark_batch_retryable(Time.now + 10.seconds, error)
+          @client.close
+          raise
         rescue Errno::ECONNREFUSED, SocketError => error
           mark_batch_retryable(Time.now + 10.seconds, error)
           raise
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,7 @@ PATH
       jwt (>= 1.5.6)
       multi_json (~> 1.0)
       net-http-persistent
-      net-http2 (~> 0.14)
+      net-http2 (~> 0.18, >= 0.18.3)
       railties
       rainbow
       thor (>= 0.18.1, < 2.0)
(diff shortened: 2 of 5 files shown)


AlexTatarnikov:
Hi @benlangfeld thank you for your PR, it is exactly what I wanted to do :)

But I'm thinking about not to retry the whole batch, we have `mark_delivered` and `mark_failed` callbacks which executes per notification and marks then as delivered/failed respectively. What if we will exclude notifications which are already delivered/failed from batch on `mark_batch_retryable` ?

lib/rpush/daemon/apns2/delivery.rb
```diff
@@ -20,7 +21,11 @@ def perform
           end
 
           # Send all preprocessed requests at once
-          @client.join
+          @client.join(timeout: CLIENT_JOIN_TIMEOUT)
```
gregblake, line 24: 👍

gregblake (approved):

benlangfeld:
> Hi @benlangfeld thank you for your PR, it is exactly what I wanted to do :)
> 
> But I'm thinking about not to retry the whole batch, we have `mark_delivered` and `mark_failed` callbacks which executes per notification and marks them as delivered/failed respectively. What if we will exclude notifications which are already delivered/failed from batch on `mark_batch_retryable` ?

I'd like to do that, but I'm under serious time constraints right now. I'd like to get the timeout itself merged ASAP in order to stem the bleed. If it's less controversial, I'll remove the retry until I can make it more selective, if that will help get this merged today.

AlexTatarnikov:
> > Hi @benlangfeld thank you for your PR, it is exactly what I wanted to do :)
> > But I'm thinking about not to retry the whole batch, we have `mark_delivered` and `mark_failed` callbacks which executes per notification and marks them as delivered/failed respectively. What if we will exclude notifications which are already delivered/failed from batch on `mark_batch_retryable` ?
> 
> I'd like to do that, but I'm under serious time constraints right now. I'd like to get the timeout itself merged ASAP in order to stem the bleed. If it's less controversial, I'll remove the retry until I can make it more selective, if that will help get this merged today.

No, it is great, thank you very much for this work 👍  I just thought maybe you have arguments for or against this.

aried3r:
Hey @benlangfeld, thank you for your PR! Would you mind adding a changelog entry? If possible, but fully optional of course, could you add what the fix does and what change of behavior people can expect if they update? Especially taking @AlexTatarnikov's feedback about selective retries into account.

benlangfeld:
@aried3r The changelog says it is auto-generated. Are you sure I should add to it manually?

aried3r:
> @aried3r The changelog says it is auto-generated. Are you sure I should add to it manually?

Oh, indeed... I forgot I did that 😅. Then no, thanks! Would you mind writing it here or in the README instead? I'll merge this in the meantime.

benlangfeld:
@aried3r I included a fuller description in the description of this PR. Let me know if you'd like any further details.

aried3r:
Thank you, that should work. I'll try to release tomorrow. I'm not on a machine right now from where I can build and push a release and releasing is not yet an automated process.

PR #566, opened 2020-09-24
Drops support for EOL rubies

benlangfeld:
According to [the Ruby website](https://www.ruby-lang.org/en/downloads/branches/), these have been EOL for a while:

Ruby 2.4
status: eol
release date: 2016-12-25
EOL date: 2020-03-31

Ruby 2.3
status: eol
release date: 2015-12-25
EOL date: 2019-03-31

Dropping support for them explicitly aligns with the effort the Ruby core team are willing to invest in them, and speeds up our CI runs.


gregblake (approved):

aried3r:
Hey @benlangfeld, I'm all for always targeting the latest Ruby versions, but if there is no benefit to the code base, I'd rather not drop old Rubies just yet. Yes, they are EOL, but looking at the [rubygems stats](https://ecosystem.rubytogether.org/) 2.3 is still heavily used.

I know it would drop CI times, but while that is a benefit to us maintainers, it's not really to the users of the gem. WDYT?

benlangfeld:
@aried3r Ok.

PR #567, opened 2020-09-25
Do not retry notifications which already have been delivered/failed

AlexTatarnikov:
If an issue happens in the middle of a request, do not mark already sent/failed notifications for retry.

This is part of an error log

```ruby
I, [2020-09-25T13:08:03.210427 #78207]  INFO -- : [2020-09-25 13:08:03] [test] 1 sent to aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
[2020-09-25 13:08:03] [WARNING] [test] Will retry 1 notifications after 2020-09-25 13:08:13 due to error (SocketError, SocketError)
W, [2020-09-25T13:08:03.210714 #78207]  WARN -- : [2020-09-25 13:08:03] [WARNING] [test] Will retry 1 notifications after 2020-09-25 13:08:13 due to error (SocketError, SocketError)
D, [2020-09-25T13:08:03.211708 #78207] DEBUG -- :    (2.1ms)  SELECT COUNT(*) FROM "rpush_notifications" WHERE (processing = FALSE AND delivered = FALSE AND failed = FALSE AND (deliver_after IS NULL OR deliver_after < '2020-09-25 10:08:03.209164'))
D, [2020-09-25T13:08:03.240860 #78207] DEBUG -- :   Rpush::Client::ActiveRecord::Notification Update All (3.0ms)  UPDATE "rpush_notifications" SET processing = FALSE, delivered = TRUE, delivered_at = '2020-09-25 10:08:03.210821' WHERE "rpush_notifications"."id" = $1  [["id", 1]]
D, [2020-09-25T13:08:03.244818 #78207] DEBUG -- :   Rpush::Client::ActiveRecord::Notification Update All (2.1ms)  UPDATE "rpush_notifications" SET processing = FALSE, delivered = FALSE, delivered_at = NULL, failed = FALSE, failed_at = NULL, retries = retries + 1, deliver_after = '2020-09-25 10:08:13.210631' WHERE "rpush_notifications"."id" = $1  [["id", 1]]
[2020-09-25 13:08:03] [ERROR] SocketError, SocketError
```

You can see two updates of the same notification there. 

We have two types of "callbacks", callbacks per particular async post inside `handle_response`

```ruby
 # lib/rpush/daemon/apnsp8/delivery.rb

def ok(notification)
  log_info("#{notification.id} sent to #{notification.device_token}")
  @batch.mark_delivered(notification)
end
```

Which puts this notification inside `@delivered` array of a `@batch` when async post was completed for particular notification. And `rescue` for the whole connection inside `perform`

```ruby
 # lib/rpush/daemon/apnsp8/delivery.rb

rescue Errno::ECONNREFUSED, SocketError, HTTP2::Error::StreamLimitExceeded => error
  # TODO restart connection when StreamLimitExceeded
  mark_batch_retryable(Time.now + 10.seconds, error)
```
Which puts the whole bunch of notifications inside `@retryable` array of a `@batch`. And as result the same notification appears in both collections.

### DIFF
--- a/lib/rpush/daemon/delivery.rb
+++ b/lib/rpush/daemon/delivery.rb
@@ -20,8 +20,7 @@ module Rpush
       end
 
       def mark_batch_retryable(deliver_after, error)
-        log_warn("Will retry #{@batch.notifications.size} notifications after #{deliver_after.strftime('%Y-%m-%d %H:%M:%S')} due to error (#{error.class.name}, #{error.message})")
-        @batch.mark_all_retryable(deliver_after)
+        @batch.mark_all_retryable(deliver_after, error)
       end
 
       def mark_delivered
--- a/lib/rpush/daemon/batch.rb
+++ b/lib/rpush/daemon/batch.rb
@@ -2,6 +2,7 @@ module Rpush
   module Daemon
     class Batch
       include Reflectable
+      include Loggable
 
       attr_reader :num_processed, :notifications, :delivered, :failed, :retryable
 
@@ -27,8 +28,6 @@ module Rpush
       end
 
       def mark_retryable(notification, deliver_after)
-        return if notification.delivered || notification.failed
-
         @mutex.synchronize do
           @retryable[deliver_after] ||= []
           @retryable[deliver_after] << notification
@@ -37,10 +36,17 @@ module Rpush
         Rpush::Daemon.store.mark_retryable(notification, deliver_after, persist: false)
       end
 
-      def mark_all_retryable(deliver_after)
+      def mark_all_retryable(deliver_after, error)
+        retryable_count = 0
+
         each_notification do |notification|
+          next if notification.delivered || notification.failed
+
+          retryable_count += 1
           mark_retryable(notification, deliver_after)
         end
+
+        log_warn("Will retry #{retryable_count} of #{@notifications.size} notifications after #{deliver_after.strftime('%Y-%m-%d %H:%M:%S')} due to error (#{error.class.name}, #{error.message})")
       end
 
       def mark_delivered(notification)
@@ -54,6 +60,7 @@ module Rpush
         @mutex.synchronize do
           @delivered = @notifications
         end
+
         each_notification do |notification|
           Rpush::Daemon.store.mark_delivered(notification, Time.now, persist: false)
         end
--- a/lib/rpush/daemon/batch.rb
+++ b/lib/rpush/daemon/batch.rb
@@ -27,19 +27,19 @@ module Rpush
       end
 
       def mark_retryable(notification, deliver_after)
+        return if notification.delivered || notification.failed
+
         @mutex.synchronize do
           @retryable[deliver_after] ||= []
           @retryable[deliver_after] << notification
         end
+
         Rpush::Daemon.store.mark_retryable(notification, deliver_after, persist: false)
       end
 
       def mark_all_retryable(deliver_after)
-        @mutex.synchronize do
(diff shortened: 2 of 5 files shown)


benlangfeld (approved):
`Rpush::Daemon::Delivery#mark_batch_retryable` will still log the total batch size as being retried. I wonder if that log might be confusing; it might be interesting to log when we're skipping items in the batch because they were already delivered or failed.

Otherwise, this looks good.

AlexTatarnikov:
@benlangfeld Will it be useful if we change this log message to something like

`Will retry #{retry_count} of #{total_count} notifications after ...` ?

benlangfeld:
> @benlangfeld Will it be useful if we change this log message to something like
> 
> `Will retry 2 of 10 notifications after ...` ?

I think that would also be fine.

AlexTatarnikov:
@benlangfeld 
* Checking if there is something at all that needs to be retried
* Improved the log message

lib/rpush/daemon/delivery.rb
```diff
@@ -20,7 +20,12 @@ def mark_retryable_exponential(notification)
       end
 
       def mark_batch_retryable(deliver_after, error)
-        log_warn("Will retry #{@batch.notifications.size} notifications after #{deliver_after.strftime('%Y-%m-%d %H:%M:%S')} due to error (#{error.class.name}, #{error.message})")
+        total_count = @batch.notifications.size
+        retry_count = total_count - @batch.delivered.size - @batch.failed.size
```
benlangfeld, line 24: Ok, so now I'm a little concerned that we're duplicating the calculation of what will and will not be retried. I think we have to move the logging down to `#mark_all_retryable` in order to avoid that.

AlexTatarnikov: We don't have such calculation inside `Rpush::Daemon::Batch` so it is not a duplication or I'm missing something but still it makes some sense because it more looks like the `Rpush::Daemon::Batch` responsibility than `Rpush::Daemon::Delivery`

benlangfeld: I'm referring to https://github.com/rpush/rpush/pull/567/files#diff-805216e5e54345d16321481b20c83391R30

This is duplicating the knowledge that delivered and failed notifications won't be included. I'd prefer only one method know about that detail.

AlexTatarnikov: I need to know the exact number for the log message and if we remove such calculation, it will be implemented as kind of counter that I'll increment here during iteration through notifications https://github.com/rpush/rpush/pull/567/files#diff-805216e5e54345d16321481b20c83391R30

But as for me this incremental counter will look strange because we already know the total amount for each kind

benlangfeld: That's fine, I'd just like to keep them as close together as possible. If you don't see a good way to do that, so be it.

AlexTatarnikov:
@benlangfeld updated with counter

benlangfeld (approved):

aried3r:
Thanks for your work, both of you! Should this go out as 5.1.1 as a follow-up to #564? Or is it rather a 5.2 according to semver?

AlexTatarnikov:
It is not primarily related to #564 but still can be treated as an internal bug fix and according to semver we can use the patch version 5.1.1

PR #569, opened 2020-10-01
Reference current interface in config template

### LINKED ISSUE #526: Question about getting info from daemon
Hello, I am working on implementing this gem to send our push notifications but I am running into an issue when trying to get info from rpush running as a daemon.

The config has a comment saying that info can be retrieved by using this method:
`Rpush::Daemon::AppRunner.runners.each do |app_id, runner|`
    `    runner.app`
    `    runner.queue_size`
`end`

but this returns an error stating 
`NoMethodError (undefined method `runners' for `Rpush::Daemon::AppRunner:Class)`

I was wondering what the current method for retrieving this information is.

benlangfeld:
Fixes #526, which was broken by https://github.com/rpush/rpush/commit/f2a3c23f5b6644d602307f5f82789fcf9879f651

### DIFF
--- a/lib/generators/templates/rpush.rb
+++ b/lib/generators/templates/rpush.rb
@@ -42,10 +42,7 @@ Rpush.reflect do |on|
   # Called when a notification is queued internally for delivery.
   # The internal queue for each app runner can be inspected:
   #
-  # Rpush::Daemon::AppRunner.runners.each do |app_id, runner|
-  #   runner.app
-  #   runner.queue_size
-  # end
+  # Rpush::Daemon::AppRunner.status
   #
   # on.notification_enqueued do |notification|
   # end


lib/generators/templates/rpush.rb
```diff
@@ -42,10 +42,7 @@
   # Called when a notification is queued internally for delivery.
   # The internal queue for each app runner can be inspected:
```
aried3r, line 43: Can't say I've used this before, but since the comment says "each app runner", is there currently no way to inspect all of them? If not, what does `.status` return? I'm unfamiliar with this API.

benlangfeld: `.status` provides the best approximation to the data that I believe this old form was exposing.

https://github.com/rpush/rpush/commit/f2a3c23f5b6644d602307f5f82789fcf9879f651 removed the implementation of this example.

aried3r: I understand. Should we adapt the comment then?

benlangfeld: In what way would you like to adapt it?

aried3r: Sorry, I kind of lost track of this PR. Looking at it again, I believe my question with regards that the new comment only gets the status from _one_ runner? Or does this return the status from all runners, still?

Since the comment says, get the status from all runners.

benlangfeld: It's still all runners :)

Issue #208, opened 2015-10-13
can we support proxy configuration for dispatcher?


modx-space:
add a proxy configuration option and provide a way to use it in different dispatcher.

We need to do a monkeypatch for this.

``` ruby
module Rpush
  module Daemon
    module Dispatcher
      class Http
        def initialize(app, delivery_class, _options = {})
          @app = app
          @delivery_class = delivery_class
          proxy = "xxx"
          @http = Net::HTTP::Persistent.new('rpush', proxy)
        end

        def dispatch(payload)
          @delivery_class.new(@app, @http, payload.notification, payload.batch).perform
        end

        def cleanup
          @http.shutdown
        end
      end
    end
  end
end
```

Can we support this in rpush?
~
Thanks

diegoRodriguezAguila:
I would also like to have proxy support, as I'm behind a proxy in work and can't develop using the gem.
@shawzt how does this monkey patch work? should it work if I put something like `proxy="proxy:8080"` ?

shawzt:
@diegoRodriguezAguila 

due to GCM uses `http` as its dispatcher, please refer to [daemon-gcm](https://github.com/rpush/rpush/blob/master/lib/rpush/daemon/gcm.rb), so we just patched the [http-dispatcher](https://github.com/rpush/rpush/blob/master/lib/rpush/daemon/dispatcher/http.rb) to add proxy support. Please see the attached code above.

Fivell:
builded in proxy support sound great, now only possibility is to patch dispatchers

PR #570, opened 2020-10-01
Added ability to set proxy to http

ormus2002:


ormus2002:
Added proxy #208

lib/rpush/daemon/dispatcher/http.rb
```diff
@@ -6,6 +6,7 @@ def initialize(app, delivery_class, _options = {})
           @app = app
           @delivery_class = delivery_class
           @http = Net::HTTP::Persistent.new(name: 'rpush')
+          @http.proxy = Rpush.config.proxy if Rpush.config.proxy.present?
```
benlangfeld, line 9: This only applies to HTTP dispatchers, and does not contemplate APNSv2/p8. Could you maybe add it to those dispatchers also?

ormus2002: Sorry but I didn't use APNS, so I cannot test it

benlangfeld: Would a spec not be enough to create confidence in its functionality?

PR #571, opened 2020-10-07
Allow opting out of foreground stdout logging

benlangfeld:
By default in foreground mode logs are directed both to the logger and to stdout. If the logger goes to stdout, you can disable foreground logging to avoid duplication.


aried3r (approved):
Seems good to me! And since no behavior is changing I think this is good to go.

PR #579, opened 2020-11-18
Feature/apns2 default headers

jkraemer:
This PR adds default `apns-*` headers to `Apns2::Delivery`, in the same way they are already implemented in `Apnsp8::Delivery`. Most notably, this makes use of the `bundle_id` stored with the application instead of having to set it as a header field on every notification.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ app.name = "ios_app"
 app.certificate = File.read("/path/to/sandbox.pem")
 app.environment = "development"
 app.password = "certificate password"
+app.bundle_id = "BUNDLE ID" # the unique bundle id of the app, like com.example.appname
 app.connections = 1
 app.save!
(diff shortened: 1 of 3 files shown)


benlangfeld:
Can you provide some reference to why this is desirable?

jkraemer:
I think in most use cases the bundle id will be the same for all notifications sent through a given application instance. It certainly is in our case. Therefore, it makes sense to handle this at the application level, not having to set it on every notification.

It's also about consistency / symmetry: the `bundle_id` field is there on the `Application` instance already, and it's used for setting the `apns-topic` default header in the Apnsp8 implementation. Same goes for the other two headers - I think it's a good thing to have the same defaults here for both Apns implementations.

benlangfeld (approved):

PR #582, opened 2020-12-22
Fix more Rails 6.1 deprecation warnings

jas14:


lib/rpush/client/active_model/apns/device_token_format_validator.rb
```diff
@@ -4,7 +4,7 @@ module ActiveModel
       module Apns
         class DeviceTokenFormatValidator < ::ActiveModel::Validator
           def validate(record)
-            return if record.device_token =~ /^[a-z0-9]\w+$/i
+            return if record.device_token =~ /\A[a-z0-9]\w+\z/i
```
benlangfeld, line 7: Could you provide a reference for this change? Is this just personal preference or does something force us to do it?

jas14:
Looks like this is failing a lot of tests, sorry for the premature PR

lib/rpush/client/active_model/apns/device_token_format_validator.rb
jas14: Yes, this comes out of [ActiveModel::Validations::FormatValidator](https://github.com/rails/rails/blob/5cfd58bbfb8425ab1931c618d98b649bab059ce6/activemodel/lib/active_model/validations/format.rb#L38-L42).

jas14:
Fixed incorrect use of `ActiveRecord::Errors`

benlangfeld (approved):

PR #596, opened 2021-03-12
Fix silent APNS notifications for Apns2 and Apnsp8

shved270189:
In this PR `apns-push-type` header was added for silent notifications.
According to documentation: https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/pushing_background_updates_to_your_app#2980040 about background notifications `APNs server requires the apns-push-type field when sending push notifications to Apple Watch, and recommends it for all platforms`. Also from documentation about POST APNS requests: https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/sending_notification_requests_to_apns#2947607 `apns-push-type` header required for watchOS 6 and later; recommended for macOS, iOS, tvOS, and iPadOS.

Legacy `Rpush::Apns::Notification` works fine without this header but it will be removed 31 March. I updated our code to `Rpush::Apnsp8::Notification` and also tested `Rpush::Apns2::Notification` and background notifications did not work. I checked APNS documentation, added this fix and now it works great.

### DIFF
--- a/lib/rpush/client/active_model/apns/notification.rb
+++ b/lib/rpush/client/active_model/apns/notification.rb
@@ -52,6 +52,10 @@ module Rpush
             self.data = (data || {}).merge(CONTENT_AVAILABLE_KEY => true)
           end
 
+          def content_available?
+            (self.data || {})[CONTENT_AVAILABLE_KEY]
+          end
+
           def as_json(options = nil) # rubocop:disable Metrics/AbcSize, Metrics/PerceivedComplexity
             json = ActiveSupport::OrderedHash.new
 
--- a/lib/rpush/daemon/apns2/delivery.rb
+++ b/lib/rpush/daemon/apns2/delivery.rb
@@ -112,6 +112,7 @@ module Rpush
           headers['apns-expiration'] = '0'
           headers['apns-priority'] = '10'
           headers['apns-topic'] = @app.bundle_id
+          headers['apns-push-type'] = 'background' if notification.content_available?
 
           headers.merge notification_data(notification)[HTTP2_HEADERS_KEY] || {}
         end
--- a/lib/rpush/daemon/apnsp8/delivery.rb
+++ b/lib/rpush/daemon/apnsp8/delivery.rb
@@ -144,6 +144,7 @@ module Rpush
           headers['apns-priority'] = '10'
           headers['apns-topic'] = @app.bundle_id
           headers['authorization'] = "bearer #{jwt_token}"
+          headers['apns-push-type'] = 'background' if notification.content_available?
 
           headers.merge notification_data(notification)[HTTP2_HEADERS_KEY] || {}
         end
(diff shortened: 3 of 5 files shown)


benlangfeld:
Could you maybe provide some detail/references on this?

shved270189:
> Could you maybe provide some detail/references on this?

@benlangfeld I updated PR description. Sorry for that :)

benlangfeld (approved):

aried3r:
Before I release this I want to make sure I understood correctly, we are fine hardcoding the value to `background`, since _only_ background notifications will have `content_available` set, correct?

The first link also states (emphasis mine):

> Additionally, the notification’s POST request should contain the apns-push-type header field with a value of background, **and the `apns-priority` field with a value of 5**.

while the second link has this to say about `apns-priority`

> The priority of the notification. If you omit this header, APNs sets the notification priority to 10.
>
> Specify 10 to send the notification immediately.
>
> Specify 5 to send the notification based on power considerations on the user’s device.

So I assume the value of `10` is a good default but should be `5` in the case of `content_available` set to `true`/`1`? Is this true for every scenario? Should we change anything here?

benlangfeld:
@gregblake @lucasfeijo Are you able to answer @aried3r's question above?

lucasfeijo:
@aried3r 

> Before I release this I want to make sure I understood correctly, we are fine hardcoding the value to background, since only background notifications will have content_available set, correct?

Yes, content_available is only supposed to be used to trigger code while your app runs in the background. [source](https://stackoverflow.com/a/65777591/1346618)

I've never worked with the apns_priority field, though. I don't wanna say something wrong.

clemenshelm:
@lucasfeijo you didn't answer the question correctly. The question was if only background notifications will have `content_available` set. And the answer to that is **NO**.

It's perfectly possible to specify an `alert` in a push notification and update the app in the background with some data in the same message, like 

```json
{
  "aps": {
    "alert": "My alert",
    "content-available": 1
  },
  "custom": "data"
}
```

This change introduces a major bug since all push notifications with `content_available: true` aren't displayed anymore all of a sudden. Therefore I'd encourage you to roll back the change.

It's perfectly possible to set the `"apns-push-type"` header manually by setting `notification.data[:headers] = { "apns-push-type" => "background" }`, so in my opinion there's no real reason to perform some intransparent magic which breaks proper existing push notifications.

clemenshelm:
For those experiencing the bug, currently the workaround is to set `notification.data[:headers] = { "apns-push-type" => "alert" }` explicitly.

lordofthedanse:
Can anyone confirm that Are background updates for Apnsp8 are working at all? I’ve been banging my head for a while on this ...

benlangfeld:
@clemenshelm @thejspr @lucasfeijo @aried3r @shved270189 It sounds like the best option here is to revert this change and either let the user set the `apns-push-type` header explicitly or provide some slightly improved API wrapper around that. Would you all vote on that option with a thumbs up or down on this message? I can handle the revert if that's what we end up choosing, but I'd ask someone else to prep an improved API if it's necessary.

PR #597, opened 2021-03-12
Drop support for Rails 5.0 and 5.1

### LINKED ISSUE #589: Drop Rails 5.0 and 5.1
Rails 5.0.0 was released June 30, 2016.
Rails 5.1.0 was released April 27, 2017.

Only Rails 5.2 receives severe security updates as of now. https://guides.rubyonrails.org/maintenance_policy.html

I'm okay with maintaining 5.2 for a little while longer if effort doesn't become to high, but older Rails versions don't support newer Ruby versions and I'd like to keep the CI matrix simple.

I'm not going to drop it right now, so people who are interested can comment, but otherwise I'd drop these two versions for v6.0.0.

ericsaupe:
Closes #589


benlangfeld:
https://github.com/rpush/rpush/blob/master/.travis.yml#L27-L28 also

benlangfeld (approved):

aried3r:
@ericsaupe, thanks for the PR! I saw this too late. Would you be up for looking for code like this one and remove it as well?

https://github.com/rpush/rpush/blob/7eb29675953b755a552c8a9f419710ddfdeb2a4c/lib/generators/templates/rename_rapns_to_rpush.rb#L21-L29

ericsaupe:
@aried3r sure!

aried3r:
Thank you, much appreciated!

PR #599, opened 2021-03-15
Remove references and checks for unsupported versions of Rails

ericsaupe:
Additional changes from #597


ericsaupe:
@aried3r here it is. Let me know if I missed anything.

aried3r:
LGTM, tests pass, I see no deprecation warnings in the logs and I can't think of anything else right now. I'll wait a bit in case @benlangfeld has a comment but from my side things are ready to go.

aried3r (approved):

benlangfeld (approved):

PR #608, opened 2021-04-13
Fixed infinite loop issue with Apnsp8 delivery

### LINKED ISSUE #606: Rpush::Daemon::Apnsp8::Delivery leaves notifications permanently in "processing"
**Describe the bug**
In `Rpush::Daemon::Apnsp8::Delivery#prepare_async_post`, when `@first_push` is `false`, it calls `#delayed_push_async`, which checks for `#streams_available?` , and loops until that is `true`. `#streams_available?` checks that `#remote_max_concurrent_streams` is greater than the current stream count. But, `#remote_max_concurrent_streams` has a branch that says, if `@client.remote_settings[:settings_max_concurrent_streams] == 0x7fffffff` then return `0`. And in that scenario, `#streams_available?` is always false, and it will loop forever, leaving the batch's notifications in "processing" permanently.

**To Reproduce**
Configure rpush to use `Apnsp8`. An instance of `Rpush::Daemon::Apnsp8::Delivery` will always process the first notification of batch, set the `@first_push` flag to `false`, and then leave all other notifications in the batch in "processing"

**Expected behavior**
It should process process all notifications. Should there be a reasonable default, like 100, in the case where `settings_max_concurrent_streams` is the default value?

**System configuration (please complete the following information):**
 - OS: macOS
 - OS version: 11.2.3
 - Ruby version: 2.4.5
 - Rails version: 4.2.11
 - Rpush version: 3.2.4 (But I've verified the same code exists on master)

**Additional context**
Thanks!

diminish7:
Fixes #606 

With Apnsp8 delivery, the second notification in a batch was always caught in an infinite loop, leaving it and all subsequent notification stuck in "processing"

### DIFF
--- a/lib/rpush/daemon/apnsp8/delivery.rb
+++ b/lib/rpush/daemon/apnsp8/delivery.rb
@@ -86,6 +86,10 @@ module Rpush
         def remote_max_concurrent_streams
           # 0x7fffffff is the default value from http-2 gem (2^31)
           if @client.remote_settings[:settings_max_concurrent_streams] == 0x7fffffff
+            # Ideally we'd fall back to `#local_settings` here, but `NetHttp2::Client`
+            # doesn't expose that attr from the `HTTP2::Client` it wraps. Instead, we
+            # chose a hard-coded value matching the default local setting from the
+            # `HTTP2::Client` class
             DEFAULT_MAX_CONCURRENT_STREAMS
           else
             @client.remote_settings[:settings_max_concurrent_streams]
--- a/lib/rpush/daemon/apnsp8/delivery.rb
+++ b/lib/rpush/daemon/apnsp8/delivery.rb
@@ -8,6 +8,7 @@ module Rpush
       class Delivery < Rpush::Daemon::Delivery
         RETRYABLE_CODES = [ 429, 500, 503 ]
         CLIENT_JOIN_TIMEOUT = 60
+        DEFAULT_MAX_CONCURRENT_STREAMS = 100
 
         def initialize(app, http2_client, token_provider, batch)
           @app = app
@@ -85,7 +86,7 @@ module Rpush
         def remote_max_concurrent_streams
           # 0x7fffffff is the default value from http-2 gem (2^31)
           if @client.remote_settings[:settings_max_concurrent_streams] == 0x7fffffff
-            0
+            DEFAULT_MAX_CONCURRENT_STREAMS
           else
             @client.remote_settings[:settings_max_concurrent_streams]
           end
--- a/spec/unit/daemon/apnsp8/delivery_spec.rb
+++ b/spec/unit/daemon/apnsp8/delivery_spec.rb
@@ -1,0 +1,53 @@
+require 'unit_spec_helper'
+
+describe Rpush::Daemon::Apnsp8::Delivery do
+  subject(:delivery) { described_class.new(app, http2_client, token_provider, batch) }
+
+  let(:app) { double(bundle_id: 'MY BUNDLE ID') }
+  let(:notification1) { double('Notification 1', data: {}, as_json: {}).as_null_object }
+  let(:notification2) { double('Notification 2', data: {}, as_json: {}).as_null_object }
+
+  let(:token_provider) { double(token: 'MY JWT TOKEN') }
+  let(:max_concurrent_streams) { 100 }
+  let(:remote_settings) { { settings_max_concurrent_streams: max_concurrent_streams } }
+  let(:http_request) { double(on: nil) }
+  let(:http2_client) do
+    double(
+      stream_count: 0,
+      call_async: nil,
+      join: nil,
+      prepare_request: http_request,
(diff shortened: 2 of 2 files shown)


lib/rpush/daemon/apnsp8/delivery.rb
```diff
@@ -8,6 +8,7 @@ module Apnsp8
       class Delivery < Rpush::Daemon::Delivery
         RETRYABLE_CODES = [ 429, 500, 503 ]
         CLIENT_JOIN_TIMEOUT = 60
+        DEFAULT_MAX_CONCURRENT_STREAMS = 100
```
diminish7, line 11: Not sure what a reasonable default is, but this seemed like a good place to start. I guess we could also add a new config for it? Or alternatively, use the batch size config? Thoughts?

aried3r: Maybe @mariannegru can shed some light on where and how the `0x7fffffff` was supposed to work. As per the comment in the code, it does come from the `http-2` gem:
https://github.com/igrigorik/http-2/blob/v0.11.0/lib/http/2/connection.rb#L15

But there's also:
https://github.com/igrigorik/http-2/blob/v0.11.0/lib/http/2/connection.rb#L24

So maybe 100 is indeed the "correct" value here.

diminish7: The max current streams gets read from the server, but if the server doesn't provide that header, http2 uses `0x7fffffff` as the default, which is supposed to behave like "infinity".

That default of 100 in http2 is for the local settings vs the server settings. So that's another possible approach is to look at local settings if server settings have the default of `0x7fffffff`

aried3r: > That default of 100 in http2 is for the local settings vs the server settings.

Ah, thanks for clarifying, I missed that.

Just to clarify, if the http2 gem was configured to use "infinite" connections, we internally mapped that to 0 and this resulted in the bug. So I assume 100 being infinitely less than infinity but bigger than 0, this would be okay, but I have no deep knowledge of the inner workings of this gem.

> I guess we could also add a new config for it?

As for making it configurable, I have no opinion on this, since while we use http2 at work, we don't use the http2 gem directly. I assume being able to pass your own http2 client would be nice, since this is currently a private method, if I understood correctly.

https://github.com/rpush/rpush/blob/v5.4.0/lib/rpush/daemon/dispatcher/apns_http2.rb#L33-L41

A good first step is probably getting rid of this bug and if people feel the need to configure it, discuss it then and how we'd expose it.

> So that's another possible approach is to look at local settings if server settings have the default of 0x7fffffff

I assume that would again be `100` (or more specifically, the default local value of the http2 gem), since we don't let users of Rpush configure this value. Since the http2 client isn't configurable at all, using the default values of the http2 gem actually sounds reasonable.

Did I understand this correctly?

diminish7: > Did I understand this correctly?

Yep, that all sounds right to me!

> using the default values of the http2 gem actually sounds reasonable

So just to make sure I'm understanding, you're voting for the approach to check local settings if the server settings are set to `0x7fffffff`? Or are you saying, leave it hard-coded at 100, which matches the default local settings?

aried3r: > So just to make sure I'm understanding, you're voting for the approach to check local settings if the server settings are set to `0x7fffffff`? Or are you saying, leave it hard-coded at 100, which matches the default local settings?

Good point. I meant it as a question, kind of. If we can, I'd say let's use the local settings. That way we don't have to decide this ourselves but rather trust the http2-gems defaults. WDYT?

diminish7: Yep, I'm on board with that 👍  I'll play around with it later this afternoon.

diminish7: Oh oops, I take it back, it's a bit more complicated than that...  The `NetHttp2::Client` doesn't expose `#local_settings`. It wraps an `HTTP2::Client`, which has both `#remote_settings` and `#local_settings`, but the `NetHttp2::Client` only exposes `#remote_settings` for some reason. To get to `#local_settings`, we'd need to get a change pushed through on [net-http2](https://github.com/ostinelli/net-http2).

So I guess now I'm back to voting for the simple, hard-coded `100` approach.

What does everyone think?

aried3r: If we can't access the local settings, hard-coding to 100 with a comment that basically repeats your comment ("ideally we'd use `local_settings` but... so we use 100.") is good enough for now, as it fixes this infinite loop.

diminish7: Comment added, thanks!

aried3r (approved):

benlangfeld (approved):

PR #614, opened 2021-05-06
Removes legacy APNS replaced with v2

### LINKED ISSUE #568: Remove APNSv1 support
The APNS binary interface [will cease to function next month](https://developer.apple.com/news/?id=11042019a). We should remove it from rpush. I propose then renaming `APNS2` in rpush to simply `APNS` to take its place.

Any objections @aried3r ?

benlangfeld:
The binary interface was shut down by Apple. Fixes #568.


aried3r:
Hey @benlangfeld! I realize this is still a draft PR, but I was wondering if we need to rename classes. I'd assume it would make updating just a bit harder and invalidating possible resources (blogs, articles, StackOverflow questions...) on Rpush with APNs2 a little harder to follow.

My opinion would be to leave everything as APNs2 as that's what people would perhaps look for, even though Apple doesn't (to my knowledge) use that name themselves. WDYT?

benlangfeld:
> Hey @benlangfeld! I realize this is still a draft PR, but I was wondering if we need to rename classes. I'd assume it would make updating just a bit harder and invalidating possible resources (blogs, articles, StackOverflow questions...) on Rpush with APNs2 a little harder to follow.
> 
> My opinion would be to leave everything as APNs2 as that's what people would perhaps look for, even though Apple doesn't (to my knowledge) use that name themselves. WDYT?

My thought process was around alignment with Apple terminology, as well as avoiding the question "What is Apns1?".

I can stick with the name if you think it's important, though.

benlangfeld:
It seems Travis has no intention of actually building this. I'm thinking we should switch to Github Actions.

benlangfeld:
I apologise for the noisy commits here, I'm having trouble installing nokogiri locally so depending on CI right now.

benlangfeld:
Closing in favor of https://github.com/rpush/rpush/pull/680.

Issue #10, opened 2014-02-03
Multiple databases support


ileitch:
_Issue by **[sonxurxo](https://github.com/sonxurxo)** from Thursday Nov 28, 2013 at 17:37 GMT_
_Originally opened as https://github.com/ileitch/rapns/issues/183_

---

Hi,

It would be nice if you could set somewhere the database to use by rapns.

I'm working on some projects and all of them are going to use rapns on the same machine, and my problem comes with the "only 1 instance of rapns" restriction.

My solution: I have a "master" rails app which I only use to configure and start rapns (daemon mode, e.g. "$ rapns production"). This app connects to the "rapns" database.

My other apps have their own database, and also access the rapns one. When I have to send a notification, I'm having to specify the rapns database connection with something like:

Rapns::App.establish_connection "rapns_#{Rails.env}"
Rapns::Notification.establish_connection "rapns_#{Rails.env}"

before creating the notification, to let ActiveRecord to find rapns classes.

So it would be great if you could specify somewhere the database configuration where rapns model objects live, in the initializer, for instance.

PR #618, opened 2021-07-03
feat: add support for multiple databases with ActiveRecord

n-studio:
fixes https://github.com/rpush/rpush/issues/10

### DIFF
--- a/lib/generators/templates/rpush.rb
+++ b/lib/generators/templates/rpush.rb
@@ -1,5 +1,7 @@
 Rpush.configure do |config|
 
+  # Set multiple database configuration with active_record
+  # config.database = {}
   # Supported clients are :active_record and :redis
   config.client = :active_record
 
--- a/lib/rpush/client/active_record.rb
+++ b/lib/rpush/client/active_record.rb
@@ -2,6 +2,7 @@ require 'active_record'
 
 require 'rpush/client/active_model'
 
+require 'rpush/client/active_record/base'
 require 'rpush/client/active_record/notification'
 require 'rpush/client/active_record/app'
 
--- a/lib/rpush/client/active_record/app.rb
+++ b/lib/rpush/client/active_record/app.rb
@@ -1,7 +1,7 @@
 module Rpush
   module Client
     module ActiveRecord
-      class App < ::ActiveRecord::Base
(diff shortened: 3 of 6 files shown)


lib/generators/templates/rpush.rb
```diff
@@ -1,5 +1,7 @@
 Rpush.configure do |config|
 
+  # Set multiple database configuration with active_record
+  # config.database = {}
```
benlangfeld, line 4: Please put this below `config.client`, similarly to `redis_options`. Also, I think we should make it clearer where the docs for possible values are / what the value of this is used for.

benlangfeld:
Please also include a changelog entry :)

lib/generators/templates/rpush.rb
n-studio: Unfortunately the `database` option must be set before the client is initialized. I added some comments about the possible values to use.

n-studio:
> Is there a way we can test this?
> 
> Please also include a changelog entry :)

I updated the changelog entry. I have no idea how to test it, but if you have an idea I can implement it.

lib/generators/templates/rpush.rb
benlangfeld: I don't think setting the client type is the point at which the client is initialized, is it? You're just building a config object here.

benlangfeld: Nope, you're right, we do init the client at that point. 🤦‍♂️

benlangfeld:
@aried3r How much do you care about test coverage for this, considering the difficulty in setting up multiple databases in CI?

aried3r:
I haven't yet needed to test multi-db Rails setups on CI myself, so I have to ask naively, shouldn't it be – at least for PostgreSQL – "just" another call do [`createdb`](https://github.com/rpush/rpush/blob/31ae99cf4214c2e51d5c7771742558b4cf2e74a4/.github/workflows/test.yml#L87-L93)?

I don't want to test Rails itself, but if there was even a single test that confirms that two DBs are being used, I'd appreciate it.

Could we also add a little bit of documentation to the README?

nicdal:
Any plans to merge this to master?

PR #620, opened 2021-07-07
FCM Support

mirkode:
This PR is a rough draft to handle the new Firebase Cloud Messaging protocol.

It currently only works by setting the needed authentication params as `ENV` variables.
Those are:
* `FIREBASE_PROJECT_ID`
* `GOOGLE_ACCOUNT_TYPE `
* `GOOGLE_CLIENT_ID `
* `GOOGLE_CLIENT_EMAIL `
* `GOOGLE_PRIVATE_KEY `
(please refer to `lib/rpush/daemon/fcm/delivery.rb:147`

The code is based off of the `gcm` implementation.

Also what is still missing are tests for everything `fcm` related.


aried3r:
Hey! Thanks for the PR. Are you using this code in production already or is this an extraction from production and you're not using this fork directly?

Testing push notifications is sadly a bit tricky since it's a long running task which also relies on clients and Google itself, so I'd love to hear at least a little success story. :)

Ping @wolak88, I believe you were also interested in this.

mirkode:
Hi @aried3r,

yes, I am using my fork quite reliably in production.
All FCM notifications are delivered as they are supposed to and until now I am quite happy.
And I obviously hope that Google is not touching the protocol again haha!
I needed to deliver FCM notifications quite urgently, which is why it now resulted in that rather hacky `ENV` variable solution. 

I was also thinking putting more effort in and also save the Firebase variables in the database, but in the end I was simply lacking time :)

lib/rpush/client/active_model/fcm/notification.rb
```diff
@@ -0,0 +1,83 @@
...
+          end
+
+          def as_json(options = nil) # rubocop:disable Metrics/PerceivedComplexity
+            json = {
+              'data' => data,
+              'android' => android_config,
+              'token' => device_token
+            }
+            json['content_available'] = content_available if content_available
+            json['notification'] = alert if alert
```
Joe-noh, line 44: @mirkode 

Sorry for jumping in. Thank you for implementing FCM support.

I'm trying to send push notification to iOS app with this branch, but the notification does not show up. 

```ruby
n = Rpush::Fcm::Notification.new

n.app = Rpush::Fcm::App.find_by(name: '...')
n.device_token = '***'
n.priority = 'normal'
n.notification = { title: 'Hello' }

n.save!
```

Rpush log says it sent succesfully but the `Notification` JSON does not have `message.notification` field.

```
pry(main)> Rpush::Fcm::Notification.first.as_json
=> {"message"=>
  {"data"=>nil,
   "android"=>
    {"notification"=>
      {"title"=>"Hello", "notification_priority"=>"PRIORITY_DEFAULT", "default_sound"=>true},
     "priority"=>5,
     "ttl"=>"86400s"},
   "token"=> "***"}}
```

Then I dig into this PR code, found this line. I am not sure this is the cause of the problem I have, I guess this should be:

```suggestion
            json['notification'] = notification if notification
```

or handle `alert` like https://github.com/rpush/rpush/blob/31ae99cf4214c2e51d5c7771742558b4cf2e74a4/lib/rpush/client/active_record/apns/active_record_serializable_notification.rb#L16-L32

mirkode: Hi @Joe-noh ,

Thanks for your suggestion! 
If it is working for you, I don't see any issue changing it to that.
I think I remember that I had a reason to change it to that, but I would need to look into this again, since this was quite some time ago :)

Thanks a lot

aried3r:
Once we have tests, maybe we can release this as an experimental feature so more people can submit feedback. It doesn't seem to be breaking existing functionality. However I'm a bit hesitant to add more gems. I wonder if we should have optional dependencies for this and perhaps also the webpush code?

Perhaps something like `rpush-fcm`, `rpush-webpush` gems would make sense. Anyway, that's beyond the scope of this PR.

WDYT @rpush/maintainers?

benlangfeld:
> Once we have tests, maybe we can release this as an experimental feature so more people can submit feedback. It doesn't seem to be breaking existing functionality. However I'm a bit hesitant to add more gems. I wonder if we should have optional dependencies for this and perhaps also the webpush code?
> 
> Perhaps something like `rpush-fcm`, `rpush-webpush` gems would make sense. Anyway, that's beyond the scope of this PR.
> 
> WDYT @rpush/maintainers?

I like that idea.

lib/rpush/client/active_model/fcm/notification.rb
Joe-noh: @mirkode

I forked and committed https://github.com/Joe-noh/rpush/commit/9c81ece8d6f916bee230ef10560ae1000dcab940, then successfully pushed notification to iOS app. A bit naive implementation though, just FYI.

lib/rpush/daemon/fcm/delivery.rb
```diff
@@ -0,0 +1,160 @@
...
+          end
+        end
+
+        def retry_message
+          "Notification #{@notification.id} will be retried after #{@notification.deliver_after.strftime('%Y-%m-%d %H:%M:%S')} (retry #{@notification.retries})."
+        end
+
+        def obtain_access_token
+          authorizer = ::Google::Auth::ServiceAccountCredentials.make_creds(scope: SCOPE)
+          authorizer.fetch_access_token
```
aried3r, line 136: I could only find the bang-! version of this. https://googleapis.dev/ruby/googleauth/latest/Signet/OAuth2/Client.html#fetch_access_token!-instance_method. Is there an alias for it somewhere that I'm missing?

In general, I'm wondering if this is the most efficient implementation. As far as I understand, this will always request a new token, even if it hasn't expired yet.

[This article](https://blog.mmyoji.com/posts/2019/02-16-get-fcm-access-token-in-ruby/) mentions caching it. I don't think we need to use Google's cache stores but could use `Rails.cache` for it. There's also a [`.needs_access_token?`](https://googleapis.dev/ruby/googleauth/latest/Signet/OAuth2/Client.html#needs_access_token%3F-instance_method) helper. For this we'd need at least v1.1.0 of `googleauth` according to https://github.com/googleapis/google-auth-library-ruby/pull/347.

What do you think?

Some debug logs would probably help, because I don't recall how often `Delivery` is being instantiated off the top of my head.

lib/rpush/daemon/gcm/delivery.rb
```diff
@@ -238,4 +238,4 @@ def describe
       end
     end
   end
-end
+end
```
aried3r, line 241: Please revert this change.

aried3r (changes requested):

Ankita9818:
Any updates on this PR?? When will we be merging it to the core rpush??

michaltomaszwlodarczyk:
any news?

RowanH:
I'm sure we all just got the reminder email that June '24 is the decommission date of this API. @aried3r as you appear to be one of the maintainers here what's the plan for ongoing support of Rpush library?. I keep checking back and there's no commits in the main repo, and this particular change which seems fundamentally pivotal to Rpush being viable has not had any commentary.

While it certainly seems that this is the go-to gem, it also seems like it might be not maintained going forward? As I said above very happy to help in what way we can (beer/gift cards etc!) to help show our appreciation, we're just not at the skill level that we would trust taking on core maintenance or providing an alternative pull request to resolve some of the PR feedback above. 

Also would be very useful to know if this isn't going to be maintained going forward so we can put in place alternative strategies. I completely understand this is the downside of a library becoming popular that overtime people just don't have the time to commit to it...  it works so well we don't want to ditch it :)

lib/rpush/daemon/fcm/delivery.rb
```diff
@@ -0,0 +1,160 @@
...
+        SCOPE = 'https://www.googleapis.com/auth/firebase.messaging'.freeze
+
+        def initialize(app, http, notification, batch)
+          if necessary_data_exists?
+            @app = app
+            @http = http
+            @notification = notification
+            @batch = batch
+
+            @uri = URI.parse("#{HOST}/v1/projects/#{ENV['FIREBASE_PROJECT_ID']}/messages:send")
```
viktor-shmigol, line 18: I think it's better to add firebase_project_id and other firebase credentials to the app model instead of the env variables.
There are projects which are supporting multi-project functionality such as "WhiteLabel" apps. At this point it will be an issue to support multiple apps.
Why not add credentials to the Rpush::Fcm::App in the same way as Rpush::Apnsp8::App?
 
E.g: 
```
    app = Rpush::Fcm::App.new(
      name: 'My app',
      project_id: ENV['FIREBASE_PROJECT_ID'],
      account_type: ENV['FIREBASE_ACCOUNT_TYPE'],
      client_id: ENV['FIREBASE_CLIENT_ID'],
      client_email: ENV['FIREBASE_CLIENT_EMAIL'],
      private_key: ENV['FIREBASE_PRIVATE_KEY']
    )
    app.save!


then just get these credentials from the app.
```
@uri = URI.parse("#{HOST}/v1/projects/#{app.project_id}/messages:send")

Thanks!

KevinColemanInc:
@mirkode are you still working on this? If not, can I (or someone else?) take this on? The deadline is in 5 months (June 2024).

AnilRh:
I made an effort to keep this MR alive - see #660. /cc @mirkode

lib/rpush/client/active_model/fcm/notification.rb
```diff
@@ -0,0 +1,83 @@
...
+                super(FCM_PRIORITY_NORMAL)
+              else
+                errors.add(:priority, 'must be one of either "normal" or "high"')
+            end
+          end
+
+          def as_json(options = nil) # rubocop:disable Metrics/PerceivedComplexity
+            json = {
+              'data' => data,
+              'android' => android_config,
```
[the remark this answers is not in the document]
benlangfeld: https://github.com/rpush/rpush/commit/840125aa568740f87e1e4a60f052748ddbe9c668#diff-6442eb03e61e6d11b126830bae34da5288b95c440ec14f88fa9d2ce11657a9a2R55

PR #624, opened 2021-09-21
Don't limit webpush registration keys

### LINKED ISSUE #623: Web Push Subscriptions In Chrome Include expirationTime
**Describe the bug**
Using version 93.0.4577.82 of Chrome to test the web push support results in an error `Validation failed: Registration must have :endpoint (String) and :keys (Hash) keys` because the subscription object also includes the key `expirationTime` which isn't in the [allowed keys list](https://github.com/rpush/rpush/blob/995a6524e269a9a73630407d0589b4623bead9d3/lib/rpush/client/active_model/webpush/notification.rb#L8)

**To Reproduce**
1. Register a Chrome service worker for push notifications and add the subscription to RPush
2. Attempt to create a notification
3. See error

**Expected behavior**
Most likely RPush should only be looking for the existence of the keys it needs, not limiting the subscription to ONLY the keys it expects

treyrich:
Resolves #623

### DIFF
--- a/lib/rpush/client/active_model/webpush/notification.rb
+++ b/lib/rpush/client/active_model/webpush/notification.rb
@@ -11,7 +11,7 @@ module Rpush
               return if record.registration_ids.size > 1
               reg = record.registration_ids.first
               unless reg.is_a?(Hash) &&
-                  reg.keys.sort == KEYS &&
+                  (KEYS-reg.keys).empty? &&
                   reg[:endpoint].is_a?(String) &&
                   reg[:keys].is_a?(Hash)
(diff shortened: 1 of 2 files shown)


benlangfeld (commented):
Thanks. Could there maybe be a test that covers this?

treyrich:
The best I could think of was modifying the existing tests to include the additional key that's appearing in the subscriptions that I'm creating. But honestly I think that any existing tests should be sufficient, all we're doing here is preventing the library from throwing an unnecessary error when there are additional keys in the subscription, but ultimately they're ignored so they don't really matter.

treyrich:
Is there anything left to do here? It would be great if this could be merged

benlangfeld (approved):

Issue #628, opened 2021-12-22
Rails 7 Time.now.to_s deprecation warning


culov:
Hi, thanks for this great gem! Reporting this deprecation warning I encountered when upgrading to Rails 7.

DEPRECATION WARNING: Time#to_s(:db) is deprecated. Please use Time#to_formatted_s(:db)

The warning is coming from `/lib/rpush/logger.rb` line 72:

`formatted_msg = "[#{Time.now.to_s(:db)}] "`

gregblake:
Hey @culov, thank you for reporting this. I opened a PR to resolve the deprecation warning: https://github.com/rpush/rpush/pull/630.

PR #630, opened 2021-12-27
Resolves Rails 7 Time.now.to_s deprecation warning

gregblake:
As @culov mentioned on https://github.com/rpush/rpush/issues/628, the following deprecation warning happens when upgrading to Rails 7:

`DEPRECATION WARNING: Time#to_s(:db) is deprecated. Please use Time#to_formatted_s(:db) instead. (called from log at /Users/greg/code/rpush/lib/rpush/logger.rb:72)`

This PR resolves the deprecation warning and closes https://github.com/rpush/rpush/issues/628.


aried3r:
For some time now, GitHub Actions seems to have problems with Rails 5.2 with Ruby 2.3 which cancels all the other runs. Perhaps with the next version where we'll support Rails 7, we can also drop Ruby 2.3 and keep Rails 5.2 support for a little longer.

What do you think?

gregblake:
I'm in support of doing that, @aried3r. 

ruby-lang ended all support for Ruby 2.3 [in 2019](https://www.ruby-lang.org/en/news/2019/03/31/support-of-ruby-2-3-has-ended/), so I think it's reasonable to drop support in rpush at this time. And the Rails team will continue supporting Rails 5.2 [until June 2022](https://endoflife.date/rails), so I like the idea of rpush supporting 5.2 for a bit longer.

aried3r:
Ok, I've opened #631 and finally the CI has successful runs again. My only concern is that there are apparently still a lot of Ruby 2.3 users https://stats.rubygems.org/. But for those, the old gem versions should continue to work.

I'll go forward with dropping support and wait a bit more for feedback this week before shipping a new major release.

PR #636, opened 2022-02-14
Fix deprecation warnings from the redis gem

sharang-d:
Fixes these warnings introduced in [Redis 4.6.0](https://github.com/redis/redis-rb/blob/21ec1dec064f43def1f47983431861e6491b770e/CHANGELOG.md#460)

```
Pipelining commands on a Redis instance is deprecated and will be removed in Redis 5.0.0.

redis.multi do
  redis.get("key")
end

should be replaced by

redis.multi do |pipeline|
  pipeline.get("key")
end

(called from /bundle_path/ruby/3.0.0/gems/rpush-7.0.0/lib/rpush/daemon/store/redis.rb:170:in `block in pending_notification_ids'}
```

Similar commit in sidekiq: https://github.com/mperham/sidekiq/commit/c792342753e24e6471b6e7ab79de09814c1d5573

### DIFF
--- a/lib/rpush/daemon/store/redis.rb
+++ b/lib/rpush/daemon/store/redis.rb
@@ -152,10 +152,10 @@ module Rpush
           retryable_ns = Rpush::Client::Redis::Notification.absolute_retryable_namespace
 
           Modis.with_connection do |redis|
-            retryable_results = redis.multi do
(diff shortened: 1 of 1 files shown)


aried3r:
Looks good to me, thank you!

PR #640, opened 2022-06-19
Document how to set title/subtitle/body for apns

osheroff:
I sort of just guessed at this method of getting more advanced data into the APNS body and it worked, figured it'd be better if it was explicit.


xcskier56:
@osheroff thank you for this! I think it's a really useful addition since I'd been trying to figure this out and your updates would have really helped

PR #641, opened 2022-06-21
Update embedding section of README.md

907th:
Lets change the file suggested for embedding code from `config/rpush.rb` to `config.ru`, because embedding command placed in config file (or Rails initializer) will run in Rake tasks, Rails console and Sidekiq process too, but we dan't want that!

Also I added the similar instruction to the "[Embedding API](https://github.com/rpush/rpush/wiki/Embedding-API)" Wiki page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,7 +391,8 @@ else
 end
 ```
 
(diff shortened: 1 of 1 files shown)


907th:
I'm closing stale issue.

PR #666, opened 2024-07-29
Added Support for Apns notification in FCM & Data notifications in FCM

amritanupam:
- Added APNS notification support in FCM (with some configurations like sound and apns-collapse-id added)
- Added support for data notifications in FCM, i.e., if no notification field is present then send data notifications


benlangfeld:
Superseded by https://github.com/rpush/rpush/pull/673

Issue #667, opened 2024-07-31
Rails 7.1: ArgumentError: missing keyword: :coder If no default coder is configured, a coder must be provided to `serialize`.


sjieg:
**Describe the bug**

After upgrading to Rails 7.1 with configurations, starting rails with show this error:
```
/gems/ruby-3.3.4/gems/activerecord-7.1.3.4/lib/active_record/attribute_methods/serialization.rb:208:in `serialize': missing keyword: :coder If no default coder is configured, a coder must be provided to `serialize`. (ArgumentError)
```

**To Reproduce**

1. Upgrade to Rails 7.1
2. Use rpush 7.0.1
3. See error

**Expected behavior**
Error not to happen

**Logs and other output**

Can come with a whole backtrace, but the issue lays with serialization.rb:208. I'll come with a PR to fix this.

**System configuration (please complete the following information):**
 - OS: Windows
 - OS version: 11
 - Ruby version: 3.3.4
 - Rails version: 7.1
 - Rpush version: 7.0.1

**Additional context**

https://guides.rubyonrails.org/configuring.html#config-active-record-default-column-serializer
Rails 7.1 `default_column_serializer` defaults to nil in 7.1, forcing to specify the serializer.
```
Rails.application.config.active_record.default_column_serializer = nil
```
This makes it best practise to add `coder: YAML` to every model with a serializer.

benlangfeld:
This was fixed in https://github.com/rpush/rpush/pull/675.

PR #668, opened 2024-07-31
667: Fix missing coder for Rails 7.1

sjieg:
Fixes https://github.com/rpush/rpush/issues/667

### DIFF
--- a/lib/rpush/client/active_record/notification.rb
+++ b/lib/rpush/client/active_record/notification.rb
@@ -7,8 +7,8 @@ module Rpush
 
         self.table_name = 'rpush_notifications'
 
-        serialize :registration_ids
-        serialize :url_args
+        serialize :registration_ids, coder: YAML
+        serialize :url_args, coder: YAML
 
         belongs_to :app, class_name: 'Rpush::Client::ActiveRecord::App'
 


benlangfeld:
https://github.com/rpush/rpush/pull/675 seems to show that more is required for Rails 7.1 support. Any ideas on the failing tests there?

sjieg:
> #675 seems to show that more is required for Rails 7.1 support. Any ideas on the failing tests there?

Having a quick peek at the errors I see a lot of
```
undefined method `attribute' for #<Rpush::Client::Redis::Webpush::App:
```
I would guess that instead of `thing.attribute(:name, "value")` it's now obsolete and you need to refactor to `thing.attributes(name: "value")`
But I can't pinpoint in the code where this originates from right now.

benlangfeld:
In https://github.com/rpush/rpush/pull/675, dealing with upstream Rails breakage by ignoring 7.1.4 (assuming the fix which broke it will be in 7.1.5), there are no test failures.

I'd appreciate if you would test again at that version of rpush and if the problem still exists, help me understand why it's not causing any test failures or write a test which fails without this fix.

WoutDev:
I'd love to see this fixed ASAP. Can I help with anything?

I'm actually already running Rpush (v7.0.1) on Rails 7.1.3.4 for a while now, I was not aware it was not supported. We are now noticing a lot of GCM DeprecatedApi errors and want to start using FCM with Rails 7.1 ASAP.

benlangfeld:
@WoutDev You can help by testing https://github.com/rpush/rpush/pull/675 to confirm if this PR is actually required or not.

sjieg:
Hey @WoutDev @benlangfeld I had a look at running the tests locally (WSL2) but unfortunately I ran out of time to get it to work.

Feel free to merge this PR into your more generic Rails 7.1 PR and close this one @benlangfeld 🇳🇱

benlangfeld:
@sjieg I need to figure out how to reproduce whatever failure you saw in a test to serve as justification for this change, TDD style.

WoutDev:
My hypothesis is that this is actually a railties thing and Rpush does not use it, hence no failing tests.

The default of 'nil' is defined in railties:
https://github.com/rails/rails/blob/main/railties/lib/rails/application/configuration.rb#L289

I don't think this project uses railsties, so if the default is not set then perhaps this default (YAML) is used instead:
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/attribute_methods/serialization.rb#L20

And that would explain the passing tests. 

The documentation mentioning the default of 'nil' linked in the original issue also hints at this:
> [...] The default value depends on the config.load_defaults target version:

As `load_defaults` is a railties thing.

benlangfeld:
Makes sense, thanks @WoutDev. I wonder if we should simply set the default to `nil` in our test suite to simulate Rails' behaviour.

benlangfeld:
I included a variation of this in https://github.com/rpush/rpush/pull/675 and that is now on the master branch.

PR #664, opened 2024-06-24
FCM Readme & Henridv's FCM Apple Support

SixiS:

PR #674, opened 2024-08-29
Get test suite working again

benlangfeld:
* Drop support for old Ruby/Rails
* Switch to new web-push gem

Credit for much of this goes to @SixiS and @Henridv from https://github.com/rpush/rpush/pull/664.

I'm splitting this out so we can consider the FCM Apple support separately from getting master green again.


benlangfeld:
This can't really be decoupled from those other changes.

PR #675, opened 2024-08-29
Support Rails 7.1

benlangfeld:


benlangfeld:
I think the test failures are mostly coming from Modis and will address its compatibility with Rails 7.1 first in https://github.com/rpush/modis/pull/52.

benlangfeld:
This is actually breakage in Rails itself: https://github.com/rails/rails/issues/52820

jaahoo:
In `rpush.gemspec`, the line is currently: `s.add_runtime_dependency 'activesupport', '>= 6.0', '< 7.1.0'`.
Shouldn't it be: `s.add_runtime_dependency 'activesupport', '>= 6.0', '< 7.1'`?
I have Rails 7.1.3.4, and I can't add it.
```
Could not find compatible versions

Because rails < 7.1.4 depends on activesupport = 7.1.3.4
  and rails >= 7.1.4, < 7.2.0 depends on activesupport = 7.1.4,
  rails < 7.2.0 requires activesupport = 7.1.3.4 OR = 7.1.4.
And because rails >= 7.2.0, < 7.2.1 depends on activesupport = 7.2.0
  and rails >= 7.2.1 depends on activesupport = 7.2.1,
  activesupport = 7.1.3.4 OR = 7.1.4 OR = 7.2.0 OR = 7.2.1 is required.
So, because rpush >= 9.0.0 depends on activesupport >= 6.0, < 7.1.0
  and Gemfile depends on rpush ~> 9.0,
  version solving has failed.
```

benlangfeld:
This is pending https://github.com/rpush/rpush/pull/668

WoutDev:
@benlangfeld 

If you want the tests to fail, adding this before `spec/spec_helper#20` does the trick:
```ruby
if client == :active_record
  require 'active_record'
  if ActiveRecord::Base.respond_to?(:default_column_serializer)
    # New default in Rails 7.1: https://github.com/rails/rails/pull/47422
    ActiveRecord::Base.default_column_serializer = nil
  end
end
```
(not the cleanest code)

The error seems to be thrown when eager loading rpush.

PR #676, opened 2024-09-01
improve dependencies for sqlite in CI

timdiggins:


benlangfeld:
An alternative would be https://github.com/rpush/rpush/pull/672.

PR #679, opened 2024-09-02
Test on newer rubies

benlangfeld:


benlangfeld:
I think this is effectively blocked on https://github.com/rpush/rpush/pull/680 due to instability in the test suite.

Issue #665, opened 2024-07-18
Publish version for pull request 660


aaronpkahn:
**Describe the bug**
We are looking to use this commit https://github.com/rpush/rpush/pull/660 for our FCM HTTP transition this week.  Please publish a new gem version!

**To Reproduce**
1. Compare commits on master to published gem versions

**Expected behavior**
A new gem version is published. Maybe 7.1.0?

aried3r:
@aaronpkahn, I'm planning to release a new version once #664 is merged and we get a green test run again.

If it will help, I am willing to drop support for older Rails and Ruby versions. Any particular version you need supported?

aaronpkahn:
We currently run Rails 7.1.1 and Ruby 3.2.4

aaronpkahn:
@aried3r checking in, do you need anything else from me?

justindossey:
Is this going to happen soon? I'm seeing notifications failing as Google shuts down the deprecated APIs.

requiemformemories:
I would like to ask if there have been any updates on this issue. Google is ramping down the deprecated API, and it might be completely shut down.

benlangfeld:
https://github.com/rpush/rpush/pull/673 should move this forward.

benlangfeld:
Anyone who can test with the current master branch would be of great assistance to ensure that a release will succeed.

benlangfeld:
@aried3r I think we can release this now. Could you do the honours?

aried3r:
@benlangfeld Thank you so much for all the work you put in!

I will try to have a look and publish a release this week.

ghillahill:
I think the error we are getting:

> [ERROR] Rpush::DeliveryError, Unable to deliver notification 19929, received error (Failed to deliver to all recipients. Errors: DeprecatedApi.)


is related to this state?

benlangfeld:
@ghillahill You need to provide more information. Code, more logs and rpush version.

benlangfeld:
Closing this as a duplicate of https://github.com/rpush/rpush/issues/588 just to keep the issue tracker clean.

benlangfeld:
@aried3r If you're short on time, please consider giving me access to push a release so as to stem the flow of confused complaints about neither GCM nor FCM working a-la #683 . I have work time set aside for bringing rpush up to scratch and a release is also blocking #680, #679, #672 and soon #675 .

`gem owner --add ben@langfeld.me rpush`

aried3r:
[...]

PR #680, opened 2024-09-02
Remove APNSv1 implementation

benlangfeld:
The binary interface was shut down by Apple. Fixes https://github.com/rpush/rpush/issues/568.

Also makes the test suite more reliable and faster.

Re-run of https://github.com/rpush/rpush/pull/614.


benlangfeld:
I think we should cut a release of current master as 7.1.0 for https://github.com/rpush/rpush/issues/665 before merging this and https://github.com/rpush/rpush/pull/672 to release as 8.0.0.

PR #701, opened 2024-09-17
Support older versions of ruby

thomasgallagher:


benlangfeld:
This is the opposite direction to the one we want to take this library in.

PR #710, opened 2024-10-08
Support for Rails 7.2

IronWarlord:
Changed alias_attribute code to the interpolated methods rails generates behind the scenes to resolve the issue where this generator broke in rails 7.1.4 and up


lib/rpush/client/active_model/pushy/app.rb
```diff
@@ -5,11 +5,19 @@ module Pushy
         module App
           def self.included(base)
             base.instance_eval do
-              alias_attribute :api_key, :auth_key
+              # alias_attribute :api_key, :auth_key # Removed due to a breaking change in rails 7.2. Waiting for a fix to be implemented
```
benlangfeld, line 8: The fix is already implemented and just waiting on a Rails release. I'd rather wait for that release than merge this change.

IronWarlord: Was not sure when the new rails release was going to be released or if they were just going to continue forward with rails 8.

You can close the PR

PR #711, opened 2024-10-16
Update FCM steps in README.md

nebbii:
Had some struggles setting up FCM in my repository with the given documentation, so I'm suggesting some extra steps people might look over on their initial tests. From the documentation it's not immediately clear that `data` is not where the title and body of a notification go. Leaving `notification` blank causes rpush to return a success, but nothing will actually arrive on the device. The project ID also might have changed labels overtime, as setting the "project ID" as my `firebase_project_id` instead of the "project number" also prevents notifications from showing up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ The app `environment` for any Apns* option is "development" for XCode installs, 
 #### Firebase Cloud Messaging
 
 You will need two params to make use of FCM via Rpush.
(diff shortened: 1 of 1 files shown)


J-Swift:
I also just went through debugging why this wasnt delivering. The update about `data` fixed my issue. Thanks for documenting it.

benlangfeld (approved):

Issue #607, opened 2021-04-13
 Errno::ECONNRESET, Connection reset by peer after a few minutes when using apns2


favrik:
Hello!,

have you seen anything like this before?  I've searched through the issues, and the ones that are similar are marked as "Stale".

**Describe the bug**
We recently switched to rpush 5.0.0 in order to use **apns2** due to the binary protocol coming to end of life.  I'm not sure if this is a bug, but we are getting the following error between successful IOS push notifications:
```
2021-03-30 03:37:17] [ERROR] Errno::ECONNRESET, Connection reset by peer
/usr/lib/ruby/2.5.0/openssl/buffering.rb:182:in `sysread_nonblock'
/usr/lib/ruby/2.5.0/openssl/buffering.rb:182:in `read_nonblock'
/vendor/bundle/ruby/2.5.0/gems/net-http2-0.18.4/lib/net-http2/client.rb:145:in `block in socket_loop'
/vendor/bundle/ruby/2.5.0/gems/net-http2-0.18.4/lib/net-http2/client.rb:142:in `loop'
/vendor/bundle/ruby/2.5.0/gems/net-http2-0.18.4/lib/net-http2/client.rb:142:in `socket_loop'
/vendor/bundle/ruby/2.5.0/gems/net-http2-0.18.4/lib/net-http2/client.rb:114:in `block (2 levels) in ensure_open'
 ```
 

**To Reproduce**
(note: we start rpush with `bundle exec rpush start -f -e <env>`)
1. Send a push notification to a IOS device: the notification is delivered.
2. Wait around 20 minutes
3. Attempt to send another IOS push notification: we get the error above, and the notification is not delivered

This is the code we are using to send a push notification:
```ruby
def send_ios_notification
    pn = Rpush::Apns2::Notification.new
    pn.app = apns_app
    pn.event_id = event_id
    pn.device_id = id
    pn.device_token = push_id
    pn.badge = sendings.where("status='sent'").count if type == :alert_message

    pn.alert = { title: title, body: message }
    pn.data = ios_data_payload(title, message, type, action)
              .merge(headers: { 'apns-topic' => apns_app.bundle_id })
    pn.save
end

def ios_data_payload(title, message, type, action)
    json = { type: type }
    json[:callback_id] = action[:callback_id] if action
    json[:title] = title if title.present?
    json[:body] = message if message.present?
    { json: json }
end
```

**Expected behavior**
We don't get a connection reset, or the notification is retried successfully

**Logs and other output**

[...]

PR #723, opened 2025-02-04
Fix APNs P8 HTTP2 error recovery

kjvarga:
Currently RPush uses faulty logic when determining if a notification should be retried.  It only recognizes certain responses from APNs as needing a retry.  Which means that any failure to actually deliver the notification will result in the notification being marked as failed.

Clearly, we should retry the notification unless - and until - we receive a response from APNs with a permanent failure code, or the notification reaches its maximum number of retries.

It is very common to receive `Errno::ECONNRESET: Connection reset by peer` errors from the APNs connection, and this issue has not been resolved in RPush.  See https://github.com/rpush/rpush/issues/607.  This error occurs because Apple closes the connection on their end after some period of inactivity.  It's also common to receive `OpenSSL::SSL::SSLError` errors from the HTTP2 socket.

When an error occurs in the `NetHttp2` client, it calls the error callback registered in `create_http2_client` and swallows the error.  This means that the `Delivery` class never sees it and cannot handle it.  So the notification remains in the processing state indefinitely. The threads which `NetHttp2` create have `abort_on_exception` set to `true`, so we cannot
raise the error there otherwise it will immediately terminate the whole process.  So we store it on the client and check for it after the `join` call.  If we see an error then we raise it in the `perform` method so that it can be handled appropriately by either retrying the notification or marking it as failed.

### DIFF
--- a/lib/rpush/daemon/apnsp8/delivery.rb
+++ b/lib/rpush/daemon/apnsp8/delivery.rb
@@ -25,6 +25,8 @@ module Rpush
 
           # Send all preprocessed requests at once
           @client.join(timeout: CLIENT_JOIN_TIMEOUT)
+          error = @client.check_for_error
+          raise error if error
         rescue NetHttp2::AsyncRequestTimeout => error
           mark_batch_retryable(Time.now + 10.seconds, error)
           @client.close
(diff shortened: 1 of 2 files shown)


benlangfeld:
Fixed in #734

PR #728, opened 2025-02-19
Allow proxy config

### LINKED ISSUE #208: can we support proxy configuration for dispatcher?

jaspreet-3911:
Fixes #208


benlangfeld:
Does this work for all services or just APNs? Would you mind adding a changelog and README entry?

benlangfeld (approved):

benlangfeld:
@aried3r CI is broken in a way that I can't fix it without admin on the rpush org. Could you take a look?

aried3r:
@benlangfeld, thanks for pinging me! Apparently this was because of an issue with GitHub itself according to https://github.com/orgs/community/discussions/151956#discussioncomment-12269165.

I re-ran all jobs and some jobs seem to have passed while others have failed, but at least the billing error seems to have gone away.

benlangfeld (approved):

Issue #729, opened 2025-03-14
iOS live activity support?


guruslan:
so I made it work on my own fork. but not sure how well it fits within the framework.
I'm using same pattern as for MUTABLE_CONTENT_KEY/CONTENT_AVAILABLE_KEY to pass custom live activity hash that gets put into the aps payload sent. Very simple patch - can send if there is interest?

It's a bit not intuitive how to use though still - user has to pick which token to send start/update/end. I'm not sure what kind of api change would need to happen in rpush to make them be more properly represented.

benlangfeld:
Please go ahead and open a PR and we can discuss there.

PR #730, opened 2025-03-18
Basic ios live activity support

guruslan:
#729

To make it work requires the following:
- if you want to start it from server:
   - obtain pushToStart token from mobile device.
   - send an event with `msg.live_activity = { 'event' => 'start', 'attributes-type' => ..., 'attributes' => ... }` and token = pushToStartToken (not your regular apns token).
   - mobile device obtains pushToken and sends to server.
   - send update events with `msg.live_activity = { 'event' => 'update', 'content-state' => ... }` and token=pushToken.
   - in iOS 18 there is some improved flow for this which I haven't tested and it might require bigger surgery to rpush.
- if you want to start from the app:
   - app starts it, sends pushToken to server.
   - send update/end events same as above.

official docs: https://developer.apple.com/documentation/activitykit/starting-and-updating-live-activities-with-activitykit-push-notifications

### DIFF
--- a/lib/rpush/client/active_model/apns/notification.rb
+++ b/lib/rpush/client/active_model/apns/notification.rb
@@ -52,6 +52,12 @@ module Rpush
             self.data = (data || {}).merge(CONTENT_AVAILABLE_KEY => true)
           end
 
+          LIVE_ACTIVITY_KEY = '__rpush_live_activity__'
+          def live_activity=(hash)
+            return unless hash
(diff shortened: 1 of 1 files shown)


guruslan:
could figure out how to run tests and add some if looks reasonable.

benlangfeld:
@guruslan Thanks. This would need tests, docs and a changelog entry to be merged.

dvinciguerra:
Can I help with this pull request?

It looks kinda interesting and I have a personal interest in it.

PR #738, opened 2025-06-07
fix: allow google credentials use proxy

codebyjass:


benlangfeld:
Could you please add a changelog entry and preferably a test for this?

